PR #2, opened 2018-04-25
Added initial testdata for ExoSeq

apeltzer:
This incorporates three things

- A Tar.bz2 file with the respective kit files (BED, interval files etc)
- A tiny dataset for testing (fastq.gz)
- The smallRef dataset provided by @szilvajuhos https://github.com/szilvajuhos/smallRef


MaxUlysse:
Regarding @szilvajuhos smallRef, as all files are already zipped, would it be better to have all of them instead of a zip?
And you problably don't need all files as well.

PR #3, opened 2018-04-25
Exoseq Testdatasets added

apeltzer:
- Renamed the branch to match "exoseq"
- Added missing reference data (omni etc)
- Removed ZIP file and stuff that is not required.

### DIFF
--- a/reference/LICENSE
+++ b/reference/LICENSE
@@ -1,6 +1,7 @@
 MIT License
 
 Copyright (c) 2017 Szilveszter Juhos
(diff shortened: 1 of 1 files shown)


reference/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017 Szilveszter Juhos
```
MaxUlysse: I think you can put yourself in here too ;-)

PR #7, opened 2018-08-10
Added radseq data

remiolsen:
For a future radseq pipeline


remiolsen:
No reference available. But it's a good idea, I was thinking of adding a the option for a reference based analysis later. But the data in this PR is de novo.

PR #10, opened 2018-09-17
Test data for rnafusion

matq007:
Test data for `rnafusion` taken from GFusion project.


ewels:
Nice! Do we need a reference genome? Is this for Human or something smaller?

matq007:
Yes, we need to have a reference genome. It's a Human. In the docs, they use (Homo_sapiens, hg19 version).

ewels:
Ok cool. The only problem with having this as the test data is that the CI tests will need to download the entire Human genome reference before running. This can be quite time consuming and has caused tests to hit their time limit in the past. We usually try to set up test data using small genomes where possible - eg. Yeast, Human Chr22 and so on. However, it could be that these tools will only run with a full ref, so let's go with this for now and we can always try something different if we run into problems 👍

PR #12, opened 2018-10-10
Cleanup

matq007:
Updated the human test data with better reads taken from the tool fusioncatcher.

### DIFF
--- a/docs/USE_EXISTING_DATA.md
+++ b/docs/USE_EXISTING_DATA.md
@@ -1,7 +1,0 @@
-# How to use an existing test dataset
-
-Please fill in the appropriate checklist below (delete whatever is not relevant). These are the most common things requested when adding a new test dataset.
-
- - [ ] Check [here](https://github.com/nf-core/test-datasets/branches/all) to find the branch corresponding to the test dataset you want to use
- - [ ] Specify in the test.config the path to the files from the test dataset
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,16 @@
-# test-datasets: `rnafusion`
+# ![nfcore/test-datasets](docs/images/test-datasets_logo.png)
```
apeltzer: Maybe you want to add here that this is testdata for rnafusion only in case someone is mistakenly going here without knowing how we organize things :-)

docs/ADD_NEW_DATA.md
```diff
@@ -0,0 +1,12 @@
+# How to add and use new test dataset
```
apeltzer: Do we want to have this here? I guess this is fine to be in `master` (to inform everyone how to add new data) but not really necessary in your branch?

README.md
matq007: Good catch, going to change

docs/ADD_NEW_DATA.md
matq007: ok, I've seen it other branches also, that's why I've added it. I'll remove the additional files then.

matq007:
@apeltzer can you confirm the final changes?

apeltzer:
Yes, thank you @matq007 !

PR #18, opened 2018-10-26
Updated all files for iGenomes R64-1-1 Ensembl release

drpatelh:
The chromosome identifiers in "genome.fa" are in UCSC format i.e. chrI, chrII etc. However, the chromosome identifiers in "genes.gtf" and "genes.bed" are in ENSEMBL format i.e. I, II etc. This will break any annotation programs that use the gtf file for annotation or other purposes.

I assumed that the genome assembly was R64-1-1 but when I ran samtools faidx on my local copy of that genome and the one in the chipseq test-datasets the chromosome sizes were different...so to avoid any inconsitencies I copied over the local copies of "genome.fa" and "genes.gtf". The "genes.bed" file seems to be exactly the same when I regenerated it with "gtf2bed" script.

I think it may also be worth adding extra information in the README.md regarding the genome release used in the test-dataset for traceability.


drpatelh:
Thanks Phil. Travis tests are now passing. Phew.
https://github.com/nf-core/atacseq/pull/1

PR #20, opened 2018-11-23
Quant msproteomics

glormph:
Added two files which are both (test)data and examples:
- peptide isoelectric point data
- ENSEMBL protein information
- Another small spectra sample


glormph:
Crap, PRed against master.

apeltzer:
Don#t you want to rename the pipeline first? otherwise, you'd have to rename this branch later on as well... ;-) (happened to me twice in the last couple of days... so I want to spare you the effort ;-))

PR #26, opened 2019-02-05
Added better reference and test data for the nascent txn pipeline

ignaciot:
Switching to a GRCh38 human chromosome 21-only reference, and a smaller sample limited to reads that map to chr21.


ewels:
Sounds good, but I'm a little confused by the diff:

```
raw/nascent/sra/nascentsample.sra → reference/chr21.fa
```

This looks like you renamed a `.sra` file to a fasta file? Without any changes (it's still a binary file so I think this is correct). I guess that's not intentional. Any idea what's happening here?

ignaciot:
Hey Phil,

I actually removed the original sra file, and created two new files, one under each directory (I'm not sure why the diff shows as a rename when I actually did a `git rm` for the old file, and it's not even in the same path). 

Instead of that full sra, I'm now providing a fastq file that only contains reads from chromosome 21 (`testdata/SRR4012402.chr21.fastq`), and the fasta file for the GRCh38 chr21 reference genome (`reference/chr21.fa`).

Does that make more sense?

ewels:
Yup - makes sense! Ok, so just GitHub being weird with the diff 👍

PR #49, opened 2019-04-30
Replace gitter link with slack link

maxulysse:
And also, I notice, that we don't seem to have slack links in the nf-core/tools template.
Is this normal?


apeltzer:
Hey! Actually, the current `dev` template shouldn't have gitter links in it anymore. Thanks for spotting htis

PR #53, opened 2019-05-08
Add transcriptome fasta to be able to test salmon-like tools

lpantano:
Hi,

I added the transcriptome fasta that corresponds to this reference data test to be able to add salmon in the future to the rnaseq pipeline.

Let me know if this works!


apeltzer:
Just wondered whether you can't use some of the existing data for rnaseq testing? e.g. this here? 

https://github.com/nf-core/test-datasets/tree/rnaseq/reference

If not this is of course okay, just wondered whether we need an extra file for salmon necessarily 👍

lpantano:
In theory, it has to be a trascriptome fasta file, I don't think the genomic fasta will work well with salmon or similar tools. And in the case to implement other alternatives to salmon, it should be re-usable. Another option is to implement a tool that transforms from GTF + GENOME to transcriptome, but the only one I know is the one included in tophat, and I don't know if installing tophat as a dependency for that would work. But if you know some alternative, let me know. Happy to adapt.

apeltzer:
No worries, I just wanted to know whether we _really_ need it ;-) 

Sounds reasonable to me and the file can be used by other methods too if there is some requirement for it later on .

drpatelh:
As you mentioned @lpantano `gffread` from cufflinks can do the GTF > Transcript fasta file conversion:
http://cole-trapnell-lab.github.io/cufflinks/file_formats/#the-gffread-utility

Only caveat is that the chromosome names would have to match but you would kind of expect this anyway (especially with `iGenomes` references).

I think it would make things easier to run if we offer the option to directly generate the transcript fasta file from GTF, and an option to explicitly specify the transcript fasta file if required. How did you generate your one?

This would mean adding `TopHat/cufflinks` to the environment though...

apeltzer:
Which is something I'd actually would like to not do ;-)

drpatelh:
Nope. It has its own package on `bioconda` which makes things much more promising:
https://anaconda.org/bioconda/gffread

lpantano:
Nice, this is awesome!

-- 
Lorena Pantano, P.h.D
Research Scientist
Biostatistics Department
Harvard T.H. Chan School of Public Health
https://mirtop.github.io <https://mirtop.github.io/>
https://lpantano.github.io

On May 9, 2019 at 1:40:18 PM, Harshil Patel (notifications@github.com)
wrote:

Nope. It has its own package on bioconda which makes things much more
promising:
https://anaconda.org/bioconda/gffread

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/nf-core/test-datasets/pull/53#issuecomment-490998553>,
or mute the thread

PR #58, opened 2019-06-05
Porepatrol test data: nanopore fastq reads

AnnaSyme:


AnnaSyme:
Thanks @drpatelh - have made some changes and gzipped the test data.

drpatelh:
Thanks @AnnaSyme . You can remove the `docs` directory too.

It would probably be better to report number of reads in the expected output than file sizes?

AnnaSyme:
Yes not sure about what the best thing to report is: one step is chopping adapters, so number of reads doesn't change but total number of bases or average read length would change. I don't capture this information though -- might see if I can add to the workflow. The second filtering step can both trim and delete reads, so both number of bases and number of reads can change. Will have a look ...

drpatelh:
👍 To be honest, Not sure how useful that information will be at this point...Maybe you can add it here when you have more realistic expectations in terms of the output?

drpatelh:
Let me know what you think, and we can merge this.

AnnaSyme:
I decided to add to the workflow some detail on number of reads (and number of bases) at each stage (start, after adapter trimming, and after filtering) as I think it would be useful to know (mainly if filtering has been too stringent). So I've updated the expected output here.

drpatelh:
Thank you! Looks great 👍

PR #59, opened 2019-06-06
[WIP] Add scrnaseq test data

olgabot:


apeltzer:
This should go to the `scrnaseq` branch but otherwise good 👍

apeltzer:
Can we drop the I1 file now that you found out its not necessary? Just ballast then 👍

PR #60, opened 2019-07-05
[FEATURE] added class 2 alleles file

Zethson:
Added a class 2 alleles file and renamed alleles to class_1_alleles


Zethson:
This will allow us to include class 2 alleles into our testing configuration.

PR #65, opened 2019-08-05
Add reference file for mirtop to work

lpantano:
Alternatively, this is download from mirbase, but to avoid this in the test, adding it here, that is more reliable.


lpantano:
related to https://github.com/nf-core/smrnaseq/pull/23

PR #69, opened 2019-09-02
Add gff of genes with all features

olgabot:
Create gff of genes for testing GFF conversion in rnaseq pipeline. Need `-F` to maintain all `gene_biotype` and such conversions

```zsh
(base)
 Mon  2 Sep - 11:53  ~/code/nf-core/test-datasets/reference   origin ☊ olgabot/add-invalid-gtf-lines ✔ 3☀ 
  gffread -F genes.gtf > genes.gff
```

### DIFF
--- a/reference/Makefile
+++ b/reference/Makefile
@@ -1,2 +1,0 @@
-genes.gff:
-	gffread -F --keep-exon-attrs genes.gtf > genes.gff


reference/Makefile
```diff
@@ -0,0 +1,2 @@
+genes.gff:
+	gffread -F --keep-exon-attrs genes.gtf > genes.gff
```
MaxUlysse: Do we really need the Makefile?
It's seems quite simple, and I believe the command line is almost entirely in the header of the gff.

olgabot: To me, Makefiles are much more explicit than hunting down the commands in the files created, as they can directly show how the file was created. I think as the file creation becomes more complicated it's easier to have the command in a Makefile so one can do `make genes.gff` rather than knowing that one needs to look into the `genes.gff` for the command. Makefiles are more universal and thus more portable across systems

Also compressing all reference files as in here with `make gz` is much simpler with a Makefile:
https://github.com/nf-core/test-datasets/pull/70/files#diff-5ee1740bafdb8c331a0485c54e8edbd8

olgabot: I'm happy to write a Makefile *and* a README so that there's a landing page on github that shows how to get these files and where they came from

PR #70, opened 2019-09-03
Add gzipped references

olgabot:
For testing https://github.com/nf-core/rnaseq/pull/275


apeltzer:
We should discuss the Makefile thing again: I'm not entirely against it but see some caveats (making it less obvious for people to do things, "is this now mandatory everywhere?" "does every pipeline have to have one?" etc pp, so we should aim to be consistent ideally in all pipelines. 

Doesn't documenting these commands in the README / docs suffice?

PR #73, opened 2019-09-04
rnaseq: tar.gz indices for STAR, HiSAT2, Salmon with chromI

olgabot:
Not show what's the right way to do this, but this is kinda combining both the smaller data in https://github.com/nf-core/test-datasets/pull/72 and the gzipped references in https://github.com/nf-core/test-datasets/pull/70


apeltzer:
But there are also Salmon / STAR indices in here that we don't need right?

PR #80, opened 2019-10-23
nanoseq: few fast5 reads for testing

lwratten:
Added the direct cDNA fast5 files (less than 50Mb) and updated the readme


drpatelh:
Thanks @lwratten ! I think Ive found a way to download all of the files in `testdata` in order to test the pipeline. Its a bit messy but it works! Ill merge this in now and add in samplesheet etc :+1:

PR #101, opened 2019-12-02
Improve sarek tests

maxulysse:
Add bam files to simplify tool testing for sarek


MaxUlysse:
It's true that I don't actually explain any of the file.
Except in the readme that they are part of Sarek testing reference and dataset.
Do you want me to explain every file then?

apeltzer:
Maybe just that this is for testing the intermediate step(s) in Sarek ?

MaxUlysse:
Should be an improvement ;-)

PR #121, opened 2020-03-06
add paired fastq for cutadapt test

piotr-faba-ardigen:
Added paired fastq for cutadapt test


piotr-faba-ardigen:
The paired end data was added

PR #125, opened 2020-03-10
Methylseq - Added simulared PE dataset

FelixKrueger:
10,000 PE E. coli reads with known methylation levels, non-directional to suit all possible bisulfite needs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,3 +2,58 @@
 Test data to be used for automated testing with the nf-core pipelines
 
 This branch contains test data for the [nf-core/methylseq](https://github.com/nf-core/methylseq) pipeline.
+
+The E. coli data set was generated as 10,000 reads of paired-end data, with average methylation levels of 80% in CpG context, and 10% in non-CG context.
+
+```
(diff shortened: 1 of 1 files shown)


ewels:
You're not _really_ meant to merge your own pull-requests 😉 

Whilst it's fine to have this here, the `methylseq` branch is really for the `methylseq` pipeline specifically. I think that your intention for this data is for it to be used with the new methylation DSL2 modules, in which case the data needs to be in the [`modules` branch](https://github.com/nf-core/test-datasets/tree/modules).

FelixKrueger:
I was just reading and following the instructions ... :)

But yea, I shall add them there as well.

FelixKrueger:
I have created a new PR, this time hopefully at the correct location: https://github.com/nf-core/test-datasets/pull/126

I'll leave the merging to another pair of eyes this time (even though "You're not *really* meant to merge your own pull-requests" doesn't sound as bad as "You're *really* not meant to merge your own pull-requests")

ewels:
Thanks! Yes I self-merge occasionally, I figure that it’s ok as long as you realise that you’re not meant to and feel bad about it 😅

Most repositories have rules to prevent (non-admins) from doing this. But apparently not this one.

PR #127, opened 2020-03-17
Added PE test file symlinks to trim_galore folder

FelixKrueger:
Trim Galore may also use the 10000 read RNA-seq test files in general/fastq/rna/


FelixKrueger:
Added report files for MuliQC module. I *could* merge this myself, but I *shouldn't*. Maybe someone else can do it?

ewels:
So the idea when structuring this repo was to have:

* For every module.
* `tools/<tool-name>/input`
* `tools/<tool-name>/output`
* In some cases the input files are super generic and not tool specific. In this case they could go in `general`.

So to adhere to this, please move these files around a bit:

* `general/reports` things should be in the output folders for the tools that generated them
* `tools/multiqc/test_GRCm38_bowtie2_stats.txt` etc should be in `tools/multiqc/input`. And in fact, given that MultiQC has so many inputs, it could be good to have them in `tools/multiqc/input/bowtie2`

Sorry for being strict with this. But we expect this repo to start collecting data for very many (potentially 100s) of tools. So it needs to be well organised and clear.

FelixKrueger:
Fair enough. I believe this should now all be in a more organised structure.

PR #128, opened 2020-03-18
Added E. coli indexes for Bowtie2 HISAT2 and Bismark

FelixKrueger:
Inaddition to input and output, some modules will require index files and/or folders.


ewels:
As these are tool specific, I guess they should be in the tool folder. Maybe `tools/bismark/indices`?

FelixKrueger:
Alright, moved them to `tools/toolname/indices`

PR #135, opened 2020-03-26
Change URLs to fix tests

maxulysse:
before: https://github.com/nf-core/sarek/runs/533843703
after: https://github.com/nf-core/sarek/runs/537050148

cf https://nfcore.slack.com/archives/CE6P95170/p1585137950001600


MaxUlysse:
Works as expected: https://github.com/nf-core/sarek/runs/537090920
Thanks @ggabernet

PR #164, opened 2020-06-11
Added HG19 version of 1240k bed

TCLamnidis:
Added an HG19 version of the 1240k bedfile, as per @jfy133 request, so test can be preformed when using the HG19 reference. 

Needed for benchmarking.


TCLamnidis:
Generated with:
```
cd reference/Human
gzip -cd 1240K.pos.list_hs37d5.0based.bed.gz | awk -v OFS="\t" '{$1="chr"$1; print $0}' | gzip >1240K.pos.list_HG19.0based.bed
```

PR #184, opened 2020-09-26
add strandedness to test_full

ggabernet:
Added strandedness column in the test_full samplesheet, as the test in AWS was failing otherwise.


drpatelh (approved):
Thanks @ggabernet ! This will all eventually be replaced by a proper full-sized dataset but should be good enough for you to play with for now with the AWS tests.

ggabernet:
Ah ok, I thought they were already full datasets, but anyways at least this way the pass :)

PR #195, opened 2020-10-26
Adding and updating the ampliseq branch

emnilsson:
Updated the readme so that it references nf-core slack instead of gitter, and added a read pair with known double primers.


apeltzer:
Thanks for the readme update - if @d4straub is happy with the additional data (doesn't interfere with other data already present) I think this is fine 👍

d4straub (approved):
I am fine with it! Thanks!

PR #203, opened 2020-11-09
Added more data to the test-double-primers-dataset.

emnilsson:
Increased the number of reads in the previous read pair and added an additional read pair, to ensure proper testing.


d4straub (approved):
Looks good to me!

PR #206, opened 2020-11-24
Changed the names of the test-files for the double_primer step

emnilsson:
Previously the names included an underscore instead of a dash.


d4straub (approved):
True, that can cause problems, apologies for not seeing this at the first review!

PR #210, opened 2021-01-19
Added test datasets for clipseq pipeline

amchakra:


metadata.csv
```diff
@@ -0,0 +1,3 @@
+sample_id,fastq
```
drpatelh, line 1:
```suggestion
sample,fastq
```

drpatelh: For simplicity and to fit with other pipelines.

metadata.csv
```diff
@@ -0,0 +1,3 @@
...
+exp1,https://raw.githubusercontent.com/nf-core/test-datasets/clipseq/reads/prpf8-hela-eif4a3-sirna-20190611-ju_trimmed_chr20_0_64444167.fq.gz,
```
drpatelh, line 2:
```suggestion
exp1,https://raw.githubusercontent.com/nf-core/test-datasets/clipseq/reads/prpf8-hela-eif4a3-sirna-20190611-ju_trimmed_chr20_0_64444167.fq.gz
```

drpatelh: We don't need the trailing comma? This pipeline will only support single-end data? If you are likely to add paired-end support then I would be tempted to have `sample,fastq_1,fastq_2` from the start.

PR #219, opened 2021-02-22
Add bigger test data

ggabernet:
- Adding bigger TCR test data with 10K reads


ggabernet:
sorry want to test it out already and not much to review here actually so I will merge!

PR #221, opened 2021-03-09
Reformatted, renamed and added samplesheet files

skrakau:
In preparation to the mag DSL2 conversion: 

- reformatted manifest files from `TSV` to `CSV` format and added header
- added manifest files also for tests using only short reads
- rename files according to test profiles (e.g. `test_hybrid` -> `*.hybrid.csv`)
- renamed files: `manifest*.csv` -> `samplesheet.csv`


skrakau:
Will close this. Instead of replacing the files with the old format, I will first add the new test data for DSL2: see https://github.com/nf-core/test-datasets/pull/226

PR #241, opened 2021-04-15
Moving human test data over

FriederikeHanssen:
# Draft PR for discussion

Moving human test data over from modules repo.

I applied some of the naming suggestions from @drpatelh PR. Personally, I find leaving `T` and `N` as marker easier to read than `1` and `2`. But since I am mostly using sarek this is probably highly subjective. Would be curious to hear what other's preference is.

Unfortunately, the gtf/gff3s I pulled from Ensembl are still not working for @FloWuenne, so for now I opted for using the Gencode ones in the hopes they work. However, I am not sure whether we should rather try to make this as similar to the sarscov data as possible. For now I left both in but when merging this, we can easily remove one of them.

Let me know what you think and what is missing. I will try to add in some of the missing files. If anyone has a good dataset that is also covering chr6 we could still replace it relatively easily, now that I have everything set up locally.


KevinMenden:
Hi @FriederikeHanssen , this looks really great!

Whether we lave `T`  and `N` or use `1`  or `2` instead really depends whether we want to try and normalize the different filenames between organisms as much as possible. Personally I like having those letters in there, as I could immediately guess what they represent. On the other hand, if we are going for a `yml` file that is associated with every data file, the information could be relatively easily retrieved from there.
I guess we just have to evaluate what the benefits of 100% normalized filenames are ... I'm just going to dump some of my thoughts here.
Pro:
* makes file names less complex
* slightly less risk of having something twice
* possible to easily switch between organisms (at least for some files and some tools)

Con:
* almost no idea what's in the file from looking at it
* it will always only be possible for certain files

Okay apart from that though, I think it looks good! I would like to try and get this in. I'm sure we find some issues with it then but it will be easier to debug using it. Ah one thing we might need at some point is a small description for every file, how it was created. In a way we have this by your README.md, but also given what we said yesterday, we probably should have some file-specific info too. 
:+1:

KevinMenden (approved):

drpatelh:
Looks supercalifragilisticexpialidocious now @FriederikeHanssen . Thanks a bunch!

PR #251, opened 2021-05-03
sarscov2 genome indexed to test the LAST aligner modules.

charles-plessy:
`genome.fasta` (MT192765.1) file indexed for the [LAST](https://gitlab.com/mcfrith/last) aligner with the `last/lastdb` nf-core module. See for details: https://github.com/nf-core/modules/blob/16d20a7cc44b353e62669fa26c1a35987d72c670/tests/software/last/lastdb/main.nf

I would like to use this data in the test suite of other LAST modules that will be submitted to _nf-core_: `last/train` and `last/lastal`.
This would allow us to avoid running `last/lastdb` in their test suites.  See also: https://github.com/nf-core/modules/issues/464 for a discussion of the proposed LAST modules.

Like with other aligners, a LAST "index" a collection of multiple (here 7) files which are expected to have all the same base name.


drpatelh:
Thanks @charles-plessy ! You may want to `tar.gz` the folder and submit since NF is unable to recursively download files over https. See [`KRAKEN_RUN`](https://github.com/nf-core/modules/blob/5e8624538844b57e698fa212f724bb43dffd5951/tests/software/kraken2/run/main.nf#L14) module.

PR #256, opened 2021-05-14
Example sequence alignment in MAF format using sarscov2 test data.

charles-plessy:
The SARS-CoV-2 genome MT192765.1 was aligned to the `contigs.fasta` example set of contig sequences provided in this repository, using the LAST aligner.  The resulting file, in Multiple Alignment File (MAF) format, is needed for the test suite of submodules of `last`, such as `last/mafswap`, `last/mafconvert`, or `last/split`, that we are preparing to submit to the nf-core/modules repository (see Issue https://github.com/nf-core/modules/issues/464).

While the file is small, I provide it compressed with gzip as typical MAF files can be large, therefore modules and their tests should be written to received them compressed.


drpatelh:
Why do we need double underscores @charles-plessy ? 😅

charles-plessy:
@drpatelh The name is constructed like: `queryID__targetID`, and as the IDs often contain single underscores but rarely double ones, I found it useful (in our homemade pipelines) to have a delimiter that makes it visually clear where each ID is.

drpatelh:
Ah, I see! Be awesome if we can revert to using single underscores as we have standardised everywhere else here please :)

drpatelh:
In fact, something like this would be even better https://github.com/nf-core/modules/pull/492/files?file-filters%5B%5D=.nf&file-filters%5B%5D=.yml#r634110937

charles-plessy:
@drpatelh OK I replaced the double underscore with a dot. (Sorry for the too many commits)

drpatelh (approved):
💥 Thank you!

PR #260, opened 2021-05-18
Example parameter file for the LAST aligner.

aleksandrabliznina:
This file contains a score matrix after training LAST to align sarscov2 contigs to the sarscov2 genome. It will be used for the test of the last/lastal module that we are preparing (https://github.com/nf-core/modules/issues/464).


drpatelh:
Can we rename like [this](https://github.com/nf-core/modules/pull/492/files?file-filters%5B%5D=.nf&file-filters%5B%5D=.yml#r634110937) please?

aleksandrabliznina:
Okay, done!

PR #263, opened 2021-05-28
feat: simplify csv files

maxulysse:
Trying something different with the samplesheet.csv files


maxulysse:
(works locally so far on my progress in https://github.com/nf-core/sarek/pull/379)

FriederikeHanssen:
Why are you removing gender information? It is needed for copy number calling

maxulysse:
I actually just remove it from the normal only samples.
cf https://github.com/nf-core/sarek/blob/c3714186c70e8d0e25072924bddf0b56f0f418ce/workflows/sarek.nf#L376-L412

So far, in previous sarek versions, we had one function to extract information from the TSV file for each step.
By using header in the csv file, we can directly assess if sample has gender, if it's for mappping or variant calling.
Hence simplifying this process.
I do like what Harshil did with the checksamplesheet process, but I feel like we can do all that in groovy/nextflow and not waste a process for it.

Basically my plan is to change as little as possible for users like (so switching from TSV to CSV(+header)).
But in the meantime, I feel like we can simplify some stuff for some other users as well.
I'm planning to start a discussion in https://github.com/nf-core/sarek/pull/379 (as soon as I manage to fix the CI tests (and other issues) again), so we can discuss of what we want to be as header in said file.

FriederikeHanssen:
Ah ok, I see. That makes sense. In the case of providing normal/tumor sample pairs, if the field is left empty, it will set it to 'NA' for both, right and CNV will be skipped?

maxulysse:
That's it.
But I think we can have the csv flexible.
Default status if not specified is normal -> `0`.
Default gender if not specified is `NA`.
That way, if at some point we want to introduce CNV for normal sample, the csv would just need to have the gender specified.

FriederikeHanssen (approved):
yes after our discussion in slack: I get what you mean. For everyone still wanting to provide all columns not much changes except tsv -> csv. The only processes needing the gender information are  CNV processes anyway. and for the CNV on germline samples not even the csv parsing code would need to be changed, but we would need to catch somewhere whether CNV tools are specified without gender information

PR #264, opened 2021-06-01
Add simulated SARS-CoV-2 Nanopore data

mjmansfi:
The repo's existing SARS-CoV-2 Nanopore data is based on amplicon sequencing, which unfortunately does not work well with long-read-based genome assemblers. Here, I have simulated "genomic" Nanopore reads using [DeepSimulator](https://github.com/liyu95/DeepSimulator). I tested with 3 assemblers, so these data should work with most applications.

Contains corresponding `fast5`, `fastq`, and `sequencing_summary.txt` files.

Note that the `fast5` archive file name is slightly inconsistent. The original data are named `test.sequencing_summary...`, `test.fastq...`, and `fast5.tar.gz`; here, I have named them all `test2.[etc]...` to keep these data clearly distinct from the original files.


mjmansfi:
Yikes! On review I see my `pytest_workflow` jobs are running out of memory. Might want to hold off on merging!! :sweat_smile:

mjmansfi:
These test data are a little bit too big, methinks. I will close this PR, and open a new one with a smaller dataset.

PR #269, opened 2021-06-16
Add RNA seq reads and STAR genome index

praveenraj2018:
1) Added RNAseq test data (FASTQ reads)
2) Added STAR index using the existing genome.fa
3) Modified the dbsnp/mills/gnomAD VCF files with correct contig header as well as coordinates adapted to start with 1
4) Added genome.interval_list file
5) Updated the README.md


maxulysse (approved):
LGTM

maxulysse:
Except, I will include the fastq in the fastq folder as:
`test.rna_1.fastq.gz` and `test.rna_2.fastq.gz`.
What do you think @KevinMenden ?

maxulysse (commented):
Just need to be sure of the location/name for the fastq files.
But looks good to me

praveenraj2018:
Can we merge this? Need to use these files for rnavar testing...

KevinMenden:
Sorry saw this too late, will have a look now

KevinMenden:
Okay so regarding the VCF files, it would be good to name them similar to here: https://github.com/nf-core/test-datasets/tree/modules/data/genomics/sarscov2/illumina/vcf

So `test1.vcf`, `test2.vcf` and so on.

I think it would be best to create a separate folder for the `dbsnp` thing.

And regarding the `rnaseq` files, have you actually tried using the available `fastq` files?

KevinMenden (changes requested):
I think the above needs to be addressed. Also pinging @drpatelh

praveenraj2018:
> Okay so regarding the VCF files, it would be good to name them similar to here: https://github.com/nf-core/test-datasets/tree/modules/data/genomics/sarscov2/illumina/vcf
> 
> So `test1.vcf`, `test2.vcf` and so on.
> 
> I think it would be best to create a separate folder for the `dbsnp` thing.
> 
> And regarding the `rnaseq` files, have you actually tried using the available `fastq` files?

@KevinMenden - yes I did run locally with the new rnaseq test fastq files and it showed variants. 
Not sure why your comments on renaming the vcf files? @maxulysse was talking about the naming convention of the fastq files. Something that you thought differently?

maxulysse:
my main concern was the fastq files:
```
data/genomics/homo_sapiens/rna_seq/fastq/rnaseq_test1.1.fastq.gz
data/genomics/homo_sapiens/rna_seq/fastq/rnaseq_test1.2.fastq.gz 
```
I would have them as:
```
data/genomics/homo_sapiens/illumina/fastq/test_rna_1.fastq.gz 
data/genomics/homo_sapiens/illumina/fastq/test_rna_2.fastq.gz 
```

I have no issues with the vcf as they are references.

Do we need star index?
Don't we rebuild it in rnaseq?

praveenraj2018:
The reason for keeping this format (rnaseq_test1.1.fastq.gz) is to indicate it is test file 1. In case if we include more test files in future, we need this naming convention. I noticed the same is followed in https://github.com/nf-core/test-datasets/tree/modules/data/genomics/homo_sapiens/illumina/fastq/

Kept the star index in the index/ for now. However we can rebuild it in pipeline in case if it does not exist. In a real scenario, one will use a pre-built index.

maxulysse:
For the first test it's just test.
Then it's test2 for the second

KevinMenden:
Hi @praveenraj2018 ,

the VCF files are just VCF files in the end and could be (hopefully) used by other tools as well. And we want to standardize names across organisms, so it would be good to name them in a similar fashion as we have for the `sarscov2` data, i.e. `test.vcf` ...
You can still indicate the original name in the README, which is always a good idea!

Regarding the fastq files, I think that was a misunderstanding, I meant whether you have tried the fastq files that are already uploaded, not the ones that you add? The idea is to try and get the tests running with the available data first, before adding anything extra. Have you tried this? And if yes, what was the issue?

And then as @maxulysse already suggested, I would rename the fastq files, if we really need the rnaseq files at all, to
`test.rnaseq_1.fastq.gz`, `test2.rnaseq_1.fastq.gz`, ...

This is the format used for the UMI fastq files, and we would like to keep the same format as good as possible.

One last comment: you will also need to add an entry to the main `README` file of that repo, and indicate where the data lies in the repo.

Thanks a lot for adding the test data! And sorry for being so picky about the naming :) We started off without a proper and strict naming scheme and it got messy real fast, so now we're trying to be really strict in order to keep everything nice and standardized.

Kevin

praveenraj2018:
Thanks @KevinMenden for the inputs.

Wondering whether I should put the files here: 
https://github.com/nf-core/test-datasets/tree/rnaseq (instead of _https://github.com/nf-core/test-datasets/tree/modules_)

It has currently data only for _S. cerevisiae_ now. I think, it is appropriate to put the new human RNAseq files here in https://github.com/nf-core/test-datasets/tree/rnaseq along with a new sample_sheet, so that all RNA test data in one place.

@maxulysse, @drpatelh  - what do you think?

maxulysse:
I think it makes more sense in the module branch as hopefully it'll be useful for other pipelines too

praveenraj2018:
Ok. 
The modules/ repo has been updated with fastq files renamed as suggested.

KevinMenden:
Okay thanks!

Sorry but I have to ask again - did the module work with the currently available FastQ data? We don't want to add redundant data.

Any reason you didn't update the VCF files? As I said it would be good to have them named `test` and then add a description in the README. Also the main README still has to be filled out please.

praveenraj2018:
@KevinMenden - yes, the VCF file has been placed in the vcf directory with name _test.rnaseq.vcf_ (also .gz and tbi). Hope this is okay. It is also now mentioned in the main README file.

I have not used the available FASTQ data in the module branch because, as I believe, they are gDNA data. As of now there is no human rna FASTQ in the branch, hence this new data.

KevinMenden:
Thank you, it would be nice if you could add the original names to the description in the README, so we know where the test.vcf files came from.

Yes, they are genomic data, but it is quite possible that your tool will still run without them, because the file format is the same (--> FASTQ). So it would be good to test this. So we can merge now as it might potentially be useful in the future but generally we really, really would like people to first try with the existing datasets if somehow possible. The output doesn't have to be sensible.

KevinMenden (approved):

praveenraj2018:
Thanks @KevinMenden @maxulysse for the support to add this data set. I am now able to use the new data path in my test.config for the rnavar pipeline and it worked generating variants vcf. Good to see that it's working :)

PR #272, opened 2021-06-29
Renamed to from airrflow to reval, added fields

ssnn-airr:
pcr_target_locus: IG[HIKL] or TR[ABDG]. Will be used to select the
receptor type for IgBLAST

single_cell: TRUE if single cells were isolated into separate compartments


ggabernet (approved):
Hi Susanna, looks good to me!

PR #274, opened 2021-06-30
Added a non-empty sarscov2 bed file

sidorov-si:
I'd like to add another non-empty tiny bed file (`test1.bed`) to `genomics/sarscov2/genome/bed/` to be able to test `bedtools intersect` properly. Currently, the test uses an emply `test2.bed` file which is probably not the best option for intersections. I made `test1.bed` from `test.bed` by randomly deleting rows, tweaking coordinates and renaming regions, so the new regions overlap with the ones from `test.bed`.


drpatelh:
Hi @sidorov-si ! This PR needs to be made relative to the `modules` branch.

drpatelh:
Also, I didn't realise that `test2.bed` was empty. Maybe we just replace that file instead. May need to update a couple of module tests on nf-core modules too.

drpatelh:
Fixed in https://github.com/nf-core/test-datasets/commit/940374a3cd50527a33693a1e4fd21e99e1e3921d

drpatelh:
I have just added those entries to `test2.bed` as suggested.

PR #283, opened 2021-07-19
Single end sample names identical to read file basenames

erikrikarddaniel:
Ampliseq failed when running with a sample sheet where sample names were identical to file's basenames, i.e. `sample0` read from `sample0.fastq.gz`. This PR modifies the existing, but not used in any test AFAIK, `Samplesheet_single_end.tsv` so that a fix for this problem can be tested.

### DIFF
--- a/samplesheets/Samplesheet_single_end.tsv
+++ b/samplesheets/Samplesheet_single_end.tsv
@@ -1,5 +1,5 @@
 sampleID	forwardReads
-sampleID_1a	https://github.com/nf-core/test-datasets/raw/ampliseq/testdata/1a_S103_L001_R1_001.fastq.gz
-sampleID_1	https://github.com/nf-core/test-datasets/raw/ampliseq/testdata/1_S103_L001_R1_001.fastq.gz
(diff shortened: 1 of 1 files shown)


d4straub:
Hi, I had a short look, but I am a little puzzled by all that commit history (31 commits!), what is this? Other than that it looks fine.

>  but not used in any test AFAIK, Samplesheet_single_end.tsv

Hm, that sample sheet should be used in another test. I can take care of that, could you open an issue in nf-core/ampliseq?

erikrikarddaniel:
Den ons 21 juli 2021 10:08Daniel Straub ***@***.***> skrev:

> Hi, I had a short look, but I am a little puzzled by all that commit
> history (31 commits!), what is this? Other than that it looks fine.
>
No idea and not by the computer until Monday. I'll check then.

> but not used in any test AFAIK, Samplesheet_single_end.tsv
>
> Hm, that sample sheet should be used in another test. I can take care of
> that, could you open an issue in nf-core/ampliseq?
>
I have a PR waiting, in which I added a test and "fixed" the issue. Monday.

PR #294, opened 2021-07-28
Same name of samples as read base names

erikrikarddaniel:
Update `Samplesheet_single.tsv` so sample names are identical to read file names minus `.fastq.gz`. This exposes an error in Ampliseq.


d4straub (approved):
Great!

PR #297, opened 2021-08-02
Add jupyter and rmarkdown notebooks for notebook module

grst:
cf. https://github.com/nf-core/modules/pull/617

Since the files have nothing to do with genomics, I created a directory `generic`. 
LMK if I should place them somewhere different.


grst:
sounds good!

PR #299, opened 2021-08-03
adds varskip files

bwlang:


genome/MN908947.3/primer_schemes/NEB/nCov-2019/vss1/neb_vss1.primer.bed
```diff
@@ -0,0 +1,148 @@
+MN908947.3	22	46	varskip-0317-1_1_LEFT	0	+
```
drpatelh, line 1: The fifth column (score field) is set to zero here for all entries and `nCoV-2019_1`/`nCoV-2019_2` in the other primer BED file? I think we should use integers here because that is what is specified in the BED spec e.g like [here](https://raw.githubusercontent.com/artic-network/artic-ncov2019/master/primer_schemes/nCoV-2019/V2/nCoV-2019.primer.bed). 

Just realised we should probably do the same for [the V1200 BED file](https://github.com/nf-core/test-datasets/raw/viralrecon/genome/MN908947.3/primer_schemes/artic/nCoV-2019/V1200/nCoV-2019.bed) too. Fancy making that change on this PR too?

Thanks!

drpatelh: Fixed in https://github.com/bwlang/test-datasets/commit/7a7277d610d71dfd9d2db1e522d95465561b968e

PR #319, opened 2021-08-12
adds repeat expansion catalog to modules

jemten:
Adds a minimal test file to the modules branch for testing with expansionhunter (https://github.com/nf-core/modules/pull/666)


maxulysse (approved):
LGTM

PR #320, opened 2021-08-13
Added two columns to Metadata.tsv

erikrikarddaniel:
As the title says, I added two columns with metadata. One that differs between samples, one that doesn't. I ran the standard Ampliseq test without problems. I think this is all I need for the moment.


d4straub (approved):
Looks good to me.
Both new columns should be compatible with old versions of the pipeline and should be ignored for statistical analysis (all values unique or all values identical in one column).

PR #325, opened 2021-08-19
Add test data for upcoming MaltExtract module

jfy133:
Also removes the `genome.fasta.zip` because this wasn't used for anything other than the UNZIP module, and so I will update tests  for that module to use the new MaltExtract `ncbitax_map.zip` file.

And bonus addition of previously forgotten entry in the README for the picard file


JoseEspinosa (approved):
LGTM! 🚀

PR #328, opened 2021-08-27
Add targets file for bcftools module

abhi18av:
This PR adds a targets file for testing the PR https://github.com/nf-core/modules/pull/669

The file has been generated using the following suggestion from the manual  http://samtools.github.io/bcftools/bcftools.html#common_options

```
bcftools query -f'%CHROM\t%POS\t%REF,%ALT\n' test2.vcf.gz | bgzip -c > test2.tsv.gz

```


jfy133 (commented):
1. Please update the modules README to describe the file
2. Maybe it would be worth naming the file as `test2.targets.tsv.gz` or similar to make it clear what is in the file

abhi18av:
Thanks for the review on this @jfy133 , I've accomodated the change requests.

Once this is merged, I'll finalize the PR for `nf-core/modules`

PR #332, opened 2021-09-05
Airrflow test data

ggabernet:
Adding airrflow test data


ssnn-airr:
The name change is happening! Yeah! I will start making the bcellmagic to airrflow change in the reveal parts.

PR #333, opened 2021-09-06
Add data for local module tests of scrnaseq 

fbdtemme:
This PR adds 3 intermediary files which are used for testing the local modules of the scrnaseq pipeline.
 
* tx2gene file
* transcriptome fasta
* tar.gz of salmon alevin results directory.


apeltzer:
I guess with the move to port the pipeline to DSLv2 it might not be necessary to have this here anymore - can you check if it would make sense to have these in the respective modules https://github.com/nf-core/modules and add them to the `modules` branch here in this repository instead? The pipeline is currently being migrated over to DSLv2 so it will be more useful to have these there for the respective modules instead 👍🏻

fbdtemme:
Some files are specific to the workflow and would indeed be a better fit in the modules branch. Others like the gff version of the annotation are not specific to any module and useful for testing the whole pipeline. I am currently making some more changes and will probably create a new pull request in the appropriate places later.

PR #340, opened 2021-09-20
Add gatk4 test data

GCJMackenzie:
This pull request adds test data which was discussed in the nf-core/modules Slack for the following Gatk4 modules:
- FilterMutectCalls
- CreateSomaticPanelOfNormals
- GenomicsdbImport
- LearnReadOrientationModel
- CalculateContamination

Each of which is finished other than needing test data on nf-core for their tests.

A summary table containing information about each individual file (or directory in the case of the genomicsdb workspace) can be found here: https://github.com/GCJMackenzie/markdowns/wiki/Test-data-for-nf-core-summary
Which includes: file name, file description, which file it was generated from, what gatk tool was used to make it and what the test file is needed to test.

All files were generated from files that exist already in the nf-core modules test-datasets, mostly the test and test2 recalibrated bam files. As such the test/test2 naming convention has been preserved for files that use just one test file. Since gatk modules can run in a tumour/normal paired mode, files that contain paired data are labelled "test_test2_paired*" and mutect2 output files include "mutect2_calls" to distinguish them from other vcf files. All tools used to generate the files either already exist on nf-core modules, or are in the list of modules above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,17 @@ If you cannot find suitable test data on this repository, please contact us on t
     * gatk:
       * test: Recalibration table corresponding to `test{,.umi}_{1,2}` (normal) reads
       * test2: Recalibration table corresponding to `test2{,.umi}_{1,2}` (tumor) reads
+      * test.pileups.table: Summarises normal allele counts, based on test.paired_end.recalibrated.sorted.bam, used by CalculateContamination
+      * test2.pileups.table: Summarises tumor allele counts, based on test2.paired_end.recalibrated.sorted.bam, used by CalculateContamination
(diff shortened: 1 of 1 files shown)


maxulysse:
I think it would be good to include the files summary in the README here: https://github.com/nf-core/test-datasets/blob/modules/data/genomics/homo_sapiens/README.md

GCJMackenzie:
@maxulysse Sure thing! I will add it in as soon as I can, thanks for taking a look.

maxulysse (approved):
I think this is some files that would be super useful, thanks a lot

GCJMackenzie:
@maxulysse Just wanted to say thanks for reviewing!

PR #362, opened 2021-10-17
Update Nascent Test data

edmundmiller:
Created a down sampled test dataset from [Identification of active transcriptional regulatory elements from GRO-seq data ]( https://pubmed.ncbi.nlm.nih.gov/25799441/)

- [x] Still need to add a table like [rnaseq sampling-information](https://github.com/nf-core/test-datasets/tree/rnaseq#sampling-information)

For the full tests, considering either using the full samples, or others from this study that were reanalyzed. https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE66031

- https://github.com/nf-core/nascent/pull/24


drpatelh (approved):
🚀

PR #344, opened 2021-09-27
Add bacteria test-data for nf-core/modules

rpetit3:
This PR adds some basic test-data for some bacteria specific `nf-core/modules` that don't quite work with SARS-CoV-2 test data. I think its a good start, since there are cases (e.g. vcf) where some of the existing test-data works great.  Overtime, bacteria-specific test data can be added as the need arises.

In these test data I've selected _Candidatus Portiera aleyrodidarum_ which a bacterial endosymbiont for White Flies. It has a genome size of 358,242bp, so it plays quite nicely with basic testing.

I'm currently using this for testing in Bactopia (https://github.com/bactopia/bactopia-tests), and here's example of GitHub Action (https://github.com/bactopia/bactopia/actions/runs/1256507737)

### About the data

- _bacteria/genome/*_   
These data are all based of [GCF_000292685](https://www.ncbi.nlm.nih.gov/assembly/GCF_000292685.1/)

- _bacteria/illumina/fastq/*_  
These data are based off FASTQs from [SRX1390609](https://www.ebi.ac.uk/ena/browser/view/SRX1390609?show=reads)

- _bacteria/nanopore/fastq/test.fastq.gz_  
While I could have used any Nanopore data (e.g. subsampled _E. coli_), I chose to use Nanopore data to match _C. aleyrodidarum_. The only Nanopore data for _C. aleyrodidarum_ was in the form of metagenomic data from [ERX3774356](https://www.ebi.ac.uk/ena/browser/view/ERX3774356?show=reads). I extracted reads _C. aleyrodidarum_ reads and thats the file found here. 

    Full details on how I created this are here: https://github.com/bactopia/bactopia-tests/tree/main/data/fastq/nanopore
    

Since this is adding a new group of test data to the modules test-data, your feedback and guidance is very much appreciated.

PR #364, opened 2021-10-18
Add full bacterial reference genome + sequencing data set

jfy133:
For a lot of microbial (meta)genomic tools, reconstructed genomes are evaluated on 'completeness' that is based on the number of expected (almost-) 'universal' housekeeping genes. A lot of these tools are designed primarily around bacteria, therefore I propose we need a bacterial based test-dataset _in addition_ to the SARS-CoV-2 one.

This PR proposes to adopt the currently used test-dataset from MAG, which is one of the main microbial focused nf-core pipelines. It uses a very widely used mammalian gut bacteria (_Bacteriodes fragilis_), which is relatively well characterised and therefore should have a lot of other associated test data that can be used with it. Furthermore, as a gut bacteria, it will likely be suitable for other metagenomics tools (in addition to typical genomics).

I've taken the main files, standardised the naming and added the original reference genome the synthetic sequencing reads were based upon.

This differs from https://github.com/nf-core/test-datasets/pull/344 because the genome of that taxon is so small, it doesn't pass some metagenomic binning tools (e.g. MaxBin2, which I'm working at the moment).

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,13 +193,17 @@ If you cannot find suitable test data on this repository, please contact us on t
         * toy.symm.upper.2.cool, toy.symm.upper.2.cp2.cool: test file for cooler_merge. Downloaded from [open2c/cooler](https://github.com/open2c/cooler/master/tests/data/toy.symm.upper.2.cool)
 
 * bacteroides_fragilis
+  * genome
+      * 'genome.fna.gz': NC_006347 genome downloaded from NCBI Genome           
   * illumina
     * fastq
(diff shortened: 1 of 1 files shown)


d4straub:
What about Nanopore data, that would be so important for many test. Mag has those as well, see [mag/test_data/minigut_reads.fastq.gz](https://github.com/nf-core/test-datasets/raw/mag/test_data/minigut_reads.fastq.gz). It would be nice to have @skrakau 's take on that because she made those test data and knows it best.
On another note, it might be good to test it with nf-core/bacass, it would be optimal if this dataset could be also used there. nf-core/bacass also has some test datasets, see [here](https://github.com/nf-core/test-datasets/tree/bacass).

jfy133:
The default bacass pipeline works with both the short reads and long reads :)

image

jfy133:
OK we have general agreement. I think this might still be a little experimental but if we mark it as so somehow, we can test it thoroughly in the hackathon? @drpatelh what do ou think (as the architect)? Quite a few modules in the hackathon will need this dataset I think.

PR #377, opened 2021-10-28
Uncompressed bacteria genome

priyanka-surana:
I am writing a BUSCO module which needs a uncompressed genome file. It doesn't work with sars-cov and human available.


jfy133:
@priyanka-surana if this is for a module, you can also include the UNZIP module to decompress the input prior to sending to BUSCO.

Alternatively, you can just allow gzipped Input, and just run `gzip -d -f $input` as a first step in the buaco module.

I've seen @grst and @maxibor do this (I'm on my phone so can't find an example atm)

priyanka-surana:
@jfy133 The busco module tests gunzip genomes. Should I delete this PR?

jfy133:
If it's unnecessary now, then yes :)

PR #389, opened 2021-11-04
Added VCF/PED pair test files

rannick:
Hi! I'd like to add a small vcf/ped pair test files to test a new module (peddy)


rannick:
Not needed anymore as the data is now part of the raredisease branch

PR #392, opened 2021-11-11
Minimal dataset to test BUSCO on human test data in `--offline` mode.

charles-plessy:
In order to run, BUSCO either downloads large reference data files, or can read from local drive when started in `--offline` mode.  The purpose of this pull request is to provide a file that allows us to test this offline mode.

BUSCO screens a genome for hits in a "lineage dataset", which have a compressed size of many megabytes.  The human sequence data in this repository (`homo_sapiens/genome/genome.fasta`) match a single entry of the primate lineage dataset (`primates_odb10.2021-02-19.tar.gz`).

In the file submitted here, I removed all the non-matching entries of the primate dataset, resulting in final size of 48 KiB.  I have verified that it can be used to test BUSCO modules in less than 2 min.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,8 @@ If you cannot find suitable test data on this repository, please contact us on t
 
 * homo_sapiens
   * genome
+    * BUSCO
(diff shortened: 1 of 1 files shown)


priyanka-surana:
@charles-plessy you also need to add a description in the README file: https://github.com/nf-core/test-datasets/blob/modules/README.md

PR #401, opened 2021-11-30
Replacing old somatic test data, adding new germline and reference data

GCJMackenzie:
Full list of files added/replaced can be found here: https://github.com/GCJMackenzie/markdowns/wiki/Test-data-replacement-addition-chart

In summarry:
This PR Replaces the bam files which were generating empty vcf files, as well as all the related gatk generated test data for my tools.
It also adds the germline and reference materials which Francesco made available alongside the somatic bam files, this includes some germline data and reference files that will be useful for the joint germline calling subworkflow.
I opted to add the reference materials as a separate subdirectory for chr21, as it means we now have two chromosomes covered, as well as a smaller and larger set of references. I do not currently use all of the germline resources so if any seem unnecessary we can remove some of those.

@lescai please let me know if you are happy with the files being uploaded, just to double check, since many of these files were originally from you.
@FriederikeHanssen if there is anything you think we are missing or could do without or needs renaming please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ If you cannot find suitable test data on this repository, please contact us on t
       * test.umi_{1,2}: UMI tagged reads corresponding to normal sample
       * test2_{1,2}: reads corresponding to tumor sample
       * test2.umi_{1,2}: UMI tagged reads corresponding to tumor sample
-      * test_{1,2}germline.fq.gz: Synthetic raw reads file used to generate normal test data for HaplotypeCaller
+      * test_{1,2}germline.fq.gz: Synthetic raw reads file used to generate normal test data for HaplotypeCaller, simulated from chr21
       * test2_{1,2}germline.fq.gz: Synthetic raw reads file used to generate disease test data for HaplotypeCaller
     * gatk:
       * test: Recalibration table corresponding to `test{,.umi}_{1,2}` (normal) reads


lescai (approved):
LGTM, happy for the use of files: they have been generated for the purpose of sharing :)

GCJMackenzie:
@lescai thanks for the confirmation!

FriederikeHanssen:
really great work, thank you so much for adding this. Just some tiny file renames:

```
1000G_omni2.5.hg38_chr21.vcf.gz -> 1000G_omni2.5.hg38.vcf.gz etc.
1000G_phase1.snps.high_confidence.hg38_chr21.vcf.gz -> 1000G_phase1.snps.hg38.vcf.gz
Axiom_Exome_Plus.genotypes.all_populations.poly.hg38_chr21.vcf.gz -> axiom_axome_plus.genotypes.all_populations.poly.hg38.vcf.gz
Mills_and_1000G_gold_standard.indels.hg38_chr21.vcf.gz -> mills_and_1000G.indels.hg38.vcf.gz
dbsnp_138.hg38_chr21.vcf.gz  -> dbsnp_138.hg38.vcf.gz  (are the multiple dbsnp files necessary?)
gnomAD.r2.1.1.GRCh38.PASS.AC.AF.only_chr21.vcf.gz -> gnomAD.r2.1.1.vcf.gz
hapmap_3.3.hg38_chr21.vcf.gz -> hapmap_3.3.hg38.vcf.gz
hapmap_3.3_grch38_pop_stratified_af_chr21.vcf.gz -> hapmap_3.pop_stratified_chr21.vcf.gz
hg38_chr21.1.bt2 -> genome.1.bt2 etc. pp
hg38_chr21.dict -> genome.dict etc for simialr files
test_haplotc_snpeff.ann.vcf.gz  -> is snpeff necessary here or can it be any annotator, in that case I would change to test_haplotc.ann.vcf.gz
```

we also would need the cram files for the bam files to make sure that works.

Other than that looks really good.

GCJMackenzie:
@FriederikeHanssen Thanks very much for the renaming suggestions! They should be implemented now. I also deleted the extra dbsnp resources and added matching cram files

FriederikeHanssen:
Awesome. ah osrry, I just remembered that there should probably be some documentation. can you add your files [here](https://github.com/nf-core/test-datasets/blob/modules/README.md) (I realized that I didn't add the cram files there my self 🤦‍♀️  and [here](https://github.com/nf-core/test-datasets/blob/modules/data/genomics/homo_sapiens/README.md). Sorry about the extra hassle, I just think it is important we document the data creation as much as possible

GCJMackenzie:
@FriederikeHanssen updated the README files, should be good to go now

README.md
```diff
@@ -140,14 +158,24 @@ If you cannot find suitable test data on this repository, please contact us on t
...
+      * test.paired_end.recalibrated.sorted: Mapped, sorted, duplicate marked, and recalibrated reads based on `test_{1,2}` (normal)
+      * test2.paired_end.sorted: Mapped, and sorted reads based on `test2_{1,2}` (tumor)
+      * test2.paired_end.markduplicates.sorted: Mapped, sorted, and duplicate marked reads based on `test2_{1,2}` (tumor)
+      * test2.paired_end.recalibrated.sorted: Mapped, sorted, duplicate marked, and recalibrated reads based on `test2_{1,2}` (tumor)
     * fastq:
       * test_{1,2}: reads corresponding to normal sample
       * test.umi_{1,2}: UMI tagged reads corresponding to normal sample
       * test2_{1,2}: reads corresponding to tumor sample
       * test2.umi_{1,2}: UMI tagged reads corresponding to tumor sample
+      * test_{1,2}germline.fq.gz: Synthetic raw reads file used to generate normal test data for HaplotypeCaller
```
FriederikeHanssen, line 177:
```suggestion
      * test_{1,2}germline.fq.gz: Synthetic raw reads file used to generate normal test data for HaplotypeCaller, simulated from chr21
```

FriederikeHanssen (approved):

PR #403, opened 2021-12-03
feat: Add first DSL2 samplesheet

jfy133:


TCLamnidis (approved):
LGTM. Would be nice to have a third sample to explicitly test 'half' UDG and BAM input validation.

PR #406, opened 2021-12-10
add pangenomics test data for the #pangenome pipeline

subwaystation:
This PR brings in test data required by https://github.com/nf-core/pangenome. 
As discussed with @heuermh in https://github.com/nf-core/pangenome/issues/74 the pangenome pipeline makes use of new data types and data structures so it requires its own set of data. 
Here is an example why the current test data is not sufficient: https://github.com/nf-core/modules/pull/1152#issuecomment-990388024. In short, the new modules can't work with the existing data, because it is malformed for the module's purpose or the data required for a tool does not exist, yet (e.g. `odgi draw` https://github.com/nf-core/modules/issues/1142).

The list of the added files with detailed explanations was also added to the README.md.

I suspect, subsequently of this PR, I have to update https://github.com/nf-core/modules/blob/master/tests/config/test_data.config, too?

Thanks for any feedback. This is my first PR here, so please be gentle. I think I followed all instructions mentioned in https://github.com/subwaystation/test-datasets/blob/modules/docs/ADD_NEW_DATA.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,7 +316,7 @@ If you cannot find suitable test data on this repository, please contact us on t
   * pangenome
     * 'pangenome.fa': A FASTA file which contains several related genomes.
     * 'pangenome.fa.gz': A GZIP compressed FASTA file which contains several related genomes.
-    * 'pangenome.paf': A GZIP compressed PAF file which contains the pairwise alignments of related genomes. 
+    * 'pangenome.paf': A PAF file which contains the all versus all pairwise alignments of related genomes. 
     * 'pangenome.paf.gz': A GZIP compressed PAF file which contains the all versus all pairwise alignments of related genomes.
     * 'pangenome.seqwish.gfa': A GFA file which contains the pangenome graph induced by `seqwish` encoded in the variation graph model.
     * 'pangenome.smoothxg.gfa': A GFA file which contains the `smoothxg` smoothed pangenome graph.


README.md
```diff
@@ -300,7 +310,22 @@ If you cannot find suitable test data on this repository, please contact us on t
       * fastq
         * 'test.fastq.gz' synthetic raw long-read sequencing reads of the genome of the mammalian-gut-residing _Bacteroides fragilis_ YCH46 bacterium (NC_006347). Originally generated for the [MAG pipeline test dataset](https://github.com/nf-core/test-datasets/tree/mag).
 
-## generic
+### pangenomics
+
+* homo_sapines
+  * pangenome
+    * 'pangenome.fa': A FASTA file which contains several related genomes for the input of `wfmash`.
```
FriederikeHanssen, line 317:
```suggestion
    * 'pangenome.fa': A FASTA file which contains several related genomes.
```
i guess other tools could use this too, right?

README.md
```diff
@@ -300,7 +310,22 @@ If you cannot find suitable test data on this repository, please contact us on t
...
+    * 'pangenome.fa.gz': A GZIP compressed FASTA file which contains several related genomes for the input of `wfmash`.
```
FriederikeHanssen, line 318:
```suggestion
    * 'pangenome.fa.gz': A GZIP compressed FASTA file which contains several related genomes.
```

README.md
```diff
@@ -300,7 +310,22 @@ If you cannot find suitable test data on this repository, please contact us on t
...
+    * 'pangenome.paf': A GZIP compressed PAF file which contains the pairwise alignments of related genomes. Ideally, it was generated by `wfmash`. It is consumed by `seqwish`.
```
FriederikeHanssen, line 319:
```suggestion
    * 'pangenome.paf': A GZIP compressed PAF file which contains the pairwise alignments of related genomes. Ideally, it was generated by `wfmash`. It is consumed by `seqwish`.
```
Don't quite understand this sentence. This PAF file isgenerated by wfmash? I would remove the "how to generate" here and add it to the other readme to keep this file as overview

README.md
```diff
@@ -300,7 +310,22 @@ If you cannot find suitable test data on this repository, please contact us on t
...
+    * 'pangenome.paf.gz': A GZIP compressed PAF file which contains the pairwise alignments of related genomes. Ideally, it was generated by `wfmash`. It is consumed by `seqwish`.
```
FriederikeHanssen, line 320: see above

README.md
```diff
@@ -300,7 +310,22 @@ If you cannot find suitable test data on this repository, please contact us on t
...
+    * 'pangenome.seqwish.gfa': A GFA file which contains the pangenome graph induced by seqwish encoded in the variation graph model. It is consumed by `smoothxg`, `odgi sort`, `odgi build`, and `odgi stats`.
```
FriederikeHanssen, line 321: see above

README.md
```diff
@@ -300,7 +310,22 @@ If you cannot find suitable test data on this repository, please contact us on t
...
+    * 'pangenome.smoothxg.gfa': A GFA file which contains the `smoothxg` smoothed pangenome graph. It is consumed by `odgi sort`, and `GFAffix`.
```
FriederikeHanssen, line 322:
```suggestion
    * 'pangenome.smoothxg.gfa': A GFA file which contains the `smoothxg` smoothed pangenome graph. 
```

README.md
```diff
@@ -300,7 +310,22 @@ If you cannot find suitable test data on this repository, please contact us on t
...
+    * 'pangenome.gfaffix.gfa': A GFA file which was normalized with `gfaffix`. It is consumed by `vg deconstruct`, `odgi layout`, `odgi draw`, and `odgi viz`.
```
FriederikeHanssen, line 323:
```suggestion
    * 'pangenome.gfaffix.gfa': A GFA file which was normalized with `gfaffix`. 
```

README.md
```diff
@@ -300,7 +310,22 @@ If you cannot find suitable test data on this repository, please contact us on t
...
+  * odgi
+    * 'pangenome.og': A variation graph encoded in the binary ODGI format. It is consumed by `odgi view`.
```
FriederikeHanssen, line 325:
```suggestion
    * 'pangenome.og': A variation graph encoded in the binary ODGI format.
```

README.md
```diff
@@ -300,7 +310,22 @@ If you cannot find suitable test data on this repository, please contact us on t
...
+    * 'pangenome.lay': A binary file which holds the 2D graph layout produced by `odgi layout`. Input for `odgi draw`.
```
FriederikeHanssen, line 326:
```suggestion
    * 'pangenome.lay': A binary file which holds the 2D graph layout produced by `odgi layout`.
```

FriederikeHanssen (changes requested):
Thank you for adding more test data. I think it would be good to have an additional README containing the commands, manual steps, etc.pp on how you cretaed all the files, just in case some updates/fixes are necessary somewhere down the road. For illumina/homo_sapiens we did it like this: https://github.com/nf-core/test-datasets/tree/modules/data/genomics/homo_sapiens not perfect, but at least something. 

And as usual the question would it be possible to reuse data that is already here, such as the sarscov2 data, something from the bacteria set or the existing human fastas?

README.md
subwaystation: I guess some of them. That depends on their downstream way of handling things.

README.md
subwaystation: Yes, such a PAF file was created by `wfmash`. 
Sorry, I should add this to which README.md?

subwaystation: FOund it!

README.md
subwaystation: This is a binary file specifically designed to be only readable by `odgi layout`. I would leave this here.

README.md
subwaystation: This is a binary file which can only be read by `odgi`. I would leave it as it is.

README.md
subwaystation: I am not sure if `seqwish` could work without having the all versus all pairwise alignments. I would suggest leaving the text here.

subwaystation:
Yes, I will add an additional README similar to https://github.com/nf-core/test-datasets/tree/modules/data/genomics/homo_sapiens. I will move the details of some files into the README.

I can't use e.g. https://github.com/nf-core/test-datasets/blob/modules/data/genomics/sarscov2/genome/genome.fasta, because it only contains one sequence entry. Non of the functionality of the modules would be tested that way, because in order to build a pangenome graph, we need a FASTA with several sequences in it. The same goes for https://github.com/nf-core/test-datasets/blob/modules/data/genomics/homo_sapiens/genome/genome.fasta.
https://github.com/nf-core/test-datasets/blob/modules/data/genomics/sarscov2/genome/proteome.fasta has several entries, but the tools can't work with protein data.

subwaystation:
I think I addressed all your concerns  @FriederikeHanssen . Could you please take a look again? Thanks!

README.md
```diff
@@ -300,7 +310,22 @@ If you cannot find suitable test data on this repository, please contact us on t
...
         * 'test.fastq.gz' synthetic raw long-read sequencing reads of the genome of the mammalian-gut-residing _Bacteroides fragilis_ YCH46 bacterium (NC_006347). Originally generated for the [MAG pipeline test dataset](https://github.com/nf-core/test-datasets/tree/mag).
 
-## generic
+### pangenomics
+
+* homo_sapines
+  * pangenome
+    * 'pangenome.fa': A FASTA file which contains several related genomes.
+    * 'pangenome.fa.gz': A GZIP compressed FASTA file which contains several related genomes.
+    * 'pangenome.paf': A GZIP compressed PAF file which contains the pairwise alignments of related genomes.
```
heuermh, line 319:
```suggestion
    * 'pangenome.paf': A PAF file which contains the all versus all pairwise alignments of related genomes. 
```

heuermh (changes requested):

README.md
subwaystation: Whoopsi, thanks!

heuermh (approved):

FriederikeHanssen (approved):
Great looks good to me now. Next step adding keys to the test.config 🚀

PR #408, opened 2021-12-15
Add sample sheet with multiple samples for full test

christopher-mohr:
This adds a sample sheet with multiple samples covering all types of input data and the two allele input formats.


ggabernet (approved):
Looks good to me

PR #411, opened 2021-12-21
Nanoseq vc

christopher-hakkaart:
Added small fastq file with known variants and a small reference genome for testing DNA variant calling functionality.


yuukiiwa (commented):
Thank you, @christopher-hakkaart!! Everything looks good!

PR #412, opened 2021-12-22
samplesheet test/data

danilodileo:
@erikrikarddaniel


erikrikarddaniel (approved):
Looks fine. (I had forgotten that we didn't have one originally, but used the #magmap file.)

PR #437, opened 2022-02-14
testdata funcscan initial commit

louperelo:


jfy133 (approved):

jfy133:
Loooks good! Once this is in please make a another PR with a test samplesheet.csv

For funcscan

```
sample,fasta
    contig_1,https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/bacteroides_fragilis/genome/genome.fna.gz
```

So please replace `contig_1` with something like `sample_1, then update the raw.github path.

We should also have a second fasta input, thinking about it :thinking: Maybe you can just copy the fasta file and give it a slightly different name for now, and insert that into the spreadsheet as e.g. `sample_2`

PR #445, opened 2022-02-23
gzip hmmer related files as hmmer accepts these

jfy133:
This cleans up `delete_me` slightly and gzips current HMMER related test files, as `hmmer` related tools accepts gzipped files (generally)


Midnighter (approved):
👍🏼

PR #455, opened 2022-02-28
add bcf test file

projectoriented:


projectoriented:
generated using this command: `bcftools view --output data/genomics/sarscov2/illumina/vcf/test.bcf --output-type b data/genomics/sarscov2/illumina/vcf/test.vcf.gz`

PR #457, opened 2022-02-28
Spatialtranscriptomics

sdomanskyi:
I have added full-test and short-test data for the spatialtranscriptomics pipeline. Each of the two datasets contains 12 files. The full test runs on 4 cores in about 20 min. The short test on the subset of data runs in 2 min. I added the script how I created the subsampled dataset.


apeltzer (approved):
Looks good to me 👍🏻

PR #464, opened 2022-03-07
added subsampled test data

jennahamlin:


jennahamlin:
Closing to add in small database and use different set of reads as test data

PR #465, opened 2022-03-09
Circdna Upload

DSchreyer:
I created a new test-data set for the circdna pipeline. These fastq files were simulated using Circle-Map and wgsim to generate sequencing reads from circular DNAs and background reads.

The test-dataset will be used in the nf-core/circdna pipeline. WIth this minimal data set, 4 out of 5 major tools can be tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,25 +39,25 @@ To track and test the reproducibility of the pipeline with default parameters be
 
 | sample	              | circles	|
 |-----------------------|-------|
-| circdna_1	| 926	  |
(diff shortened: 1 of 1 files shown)


DSchreyer:
This is the only test-data set used in the circdna pipeline. The pipeline is tested fully in its functionality by using -stub-run and by using this data set. 4 out of the 5 tools can be run using this minimal dataset and the last one is tested with stub. 

The original commits are from the nf-core pipeline atacseq. I pulled it to generate all necessary files and the file system.

PR #469, opened 2022-03-17
Change based on discussion

apeltzer:
Hi both,

this includes the proposed changes we had discussed earlier today and summarized in https://github.com/nf-core/scrnaseq/issues/92. 

Please have a look, especially not sure about the "expected_cells" bit here. For testing obviously not a super critical requirement, but should likely be chosen properly. 

We should also delete the branch this is coming from, once this is merged.


ggabernet (approved):
Looks good to me, for the `cellranger multi` workflow one would just add the necessary optional column to specify the features.

apeltzer:
Should I mention this in the readme or is this something we rather document on the pipeline level?

ggabernet:
Probably better to document that on the pipeline level...

grst:
Looks good! 
Could the example also cover a case where there are multiple files per sample? Would be clearer for the user that this is possible and would be great if that case was tested on the CI

apeltzer:
Done :-)

PR #474, opened 2022-03-18
metamaps database

henningonsbring:
This PR includes a database for metamaps


ramprasadn:
@henningonsbring DB.fa and DB_proteins.faa files seem to have the same sequence?

henningonsbring:
@ramprasadn I will look into this. DB.fa is a file made by me. The DB_proteins.faa is made by MetaMaps when their create database process in invoked.

jfy133 (approved):
@henningonsbring please update the README to describe what these files are, but otherwise looks OK to me

however please consider that EVERY one of these files will need to be described, and _every_ file be listed in test.config of the nf-core/modules repository, and each one would therefore have to be declared as an input channel.

It _may_ be easier to put them into a single directory, tar.gz archive it, and then within modules tests you can have a step that just untars the directory, and you pass a single directory as input for th eDB.

henningonsbring:
thanks for comments @ramprasadn and @jfy133 

I updated according to @jfy133 suggestion and uploaded a tar file

jfy133 (approved):

PR #489, opened 2022-03-28
Smaller size of minigut file

sofstam:
This PR has a smaller size of minigut file


jfy133:
The `fastq.gz` file isn't actually gzipped, and it still runs out of memory for me :\

If you run `fastqc -t 2 <file>` do you not also get the Java heap space out of memory error?

sofstam:
I got the same issue. I will close this PR and open a new one with our own test datasets instead.

PR #491, opened 2022-03-28
Update test dataset for long reads for centrifuge.

sofstam:
Last update for test datasets for long reads for centrifuge from [ENA](https://www.ebi.ac.uk/ena/browser/view/ERR3201942?show=reads). FastQC should work fine as well.


sofstam:
@jfy133 Is it ok if I merge this one?

PR #499, opened 2022-04-08
Add haplotype map reference

matthdsm:


FriederikeHanssen (commented):
Would it make sense to reduce this to only have the lines for chr22/21 and rename the fato genome.fasta?

matthdsm:
Done! I don't see why it wouldn't work

matthdsm:
perhaps the file would be better suited under `germline_resources` instead of `genome`

FriederikeHanssen:
yep sounds good, and a mention in the READme would also be good.

matthdsm:
Done! Who do I bother for a merge?

JoseEspinosa (approved):
LGTM after applying the changes nicely suggested by @FriederikeHanssen, thanks @matthdsm :rocket:

PR #505, opened 2022-04-12
Add kaiju to database file

sofstam:


jfy133 (changes requested):
This won't work for the test data, you need to supply a `.tar.gz` version with everything inside that we untar. This is preferred for us because then anyone who wants to  tst the pipeline and all the tools.

The reason why is we COULD add to the CI tests a download step where we download each file with `wget` and supply the directory instead. However this is not as portable as then we have to explain to a user what they have to do 'manually' to run the test profile.

I would suggest you donwload the kaiju test data that Tanja uploaded to modules, tar.gz and upload it to the taxprofiler branch instead

sofstam:
Thank you for the comments, I will update this PR :)

jfy133 (approved):
Looks good - do we need the `minigut_cf` file here, as we've already got it on `modules`?

sofstam:
I removed it :)

PR #510, opened 2022-04-13
adding kraken outputs

lescai:
Following the test data used for kraken2 module, I am adding the kraken2 outputs under the same organism.
These will be used as test data for the krona module

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,8 +130,8 @@ If you cannot find suitable test data on this repository, please contact us on t
     * sequencing_summary
       * 'test.sequencing_summary.txt'
   * taxonomy
-      * 'test_1.kraken2.reads.txt': kraken classification of each input read of test file test_1.fastq.gz
-      * 'test_1.kraken2.report.txt': kraken report after classification of test file 'test_1.fastq.gz'
+      * 'test_1.kraken2.reads.txt': kraken classification of each input read of test file `test_1.fastq.gz`
+      * 'test_1.kraken2.report.txt': kraken report after classification of test file `test_1.fastq.gz`
 
 * homo_sapiens
   * genome


FriederikeHanssen:
There is a `db` subfolder. Is this data based on the kraken data in there?

lescai:
nope it's really kraken results, i.e. processing the illumina test data, and the output that's generated. so wouldn't fit under "illumina" nor the "db".  so technically it is indeed the taxonomic classification of sars-cov-2.
but happy to move anywhere else

FriederikeHanssen:
ok, sounds sensible (I haven't used kraken yet, so you can tell me anything 😆 )

README.md
```diff
@@ -129,6 +129,9 @@ If you cannot find suitable test data on this repository, please contact us on t
       * 'test_2.fastq.gz'
     * sequencing_summary
       * 'test.sequencing_summary.txt'
+  * taxonomy
+      * 'test_1.kraken2.reads.txt': kraken classification of each input read of test file test_1.fastq.gz
```
FriederikeHanssen, line 133:
```suggestion
      * 'test_1.kraken2.reads.txt': kraken classification of each input read of test file `test_1.fastq.gz`
```

README.md
```diff
@@ -129,6 +129,9 @@ If you cannot find suitable test data on this repository, please contact us on t
...
+      * 'test_1.kraken2.report.txt': kraken report after classification of test file 'test_1.fastq.gz'
```
FriederikeHanssen, line 134:
```suggestion
      * 'test_1.kraken2.report.txt': kraken report after classification of test file `test_1.fastq.gz`
```

jfy133:
What about metagenome instead for consistency?

genome has reference files, metagenome would indicate what the whole metagenome would look like. This may also be useful as more metagenomic tools that will require downstream modules are added too?

lescai:
no problem, sounds good to me.
changes committed :)

PR #511, opened 2022-04-14
Add antismash database files

jasmezz:
During antismash database download, a couple of files are being generated within the antismash docker and singularity containers which need root permissions. The suggested workaround is to mount those files externally to the container.


jfy133 (approved):
This sucks so hard but I don't see any way around at the moment...

PR #512, opened 2022-04-19
Add fields required by Stranger

ljmesi:
Add two more fields to `repeat_expansions.json` file. These fields are required by Stranger to work properly.


ljmesi:
I noticed that review is not required for merging so I go ahead and merge this one.

PR #521, opened 2022-05-02
Adding Krona test database

lescai:
This PR adds a krona test database to the Metagenome folder of sars-cov-2.
The added files allows more realistic tests of the krona modules, following the addition of kraken reports to this repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,9 +129,10 @@ If you cannot find suitable test data on this repository, please contact us on t
       * 'test_2.fastq.gz'
     * sequencing_summary
       * 'test.sequencing_summary.txt'
-  * taxonomy
+  * metagenome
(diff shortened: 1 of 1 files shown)


jfy133 (changes requested):
Please update README, but otherwise good to go I think, file suffix could be `tsv` or similar though rather than `tab` (unless it's direct output from Krona)

lescai:
exactly :) ktUpdateTaxonomy.sh creates a file called "taxonomy.tab" so I thought best to reflect that. The ktimporttaxonomy I  believe (haven't tried though) expects a .tab file inside the folder passed as database

lescai (commented):
Should be ok now :) thanks for spotting the readme!

jfy133 (approved):

PR #524, opened 2022-05-03
feat added chrM to chr20 fasta file and removed the original

Lucpen:
Added chrM to Homo_sapiens_assembly38_chr20.fasta and renamed it to Homo_sapiens_assembly38_chr20_chrM.fasta. This was done to be able to use the sarek test-data to test the rare-disease pipeline, which includes a MT workflow.


Lucpen:
@maxulysse @ramprasadn do you think I should add any other files containing chrM (.fai, .dict...) ?

maxulysse:
@FriederikeHanssen what do you think of this PR?

Lucpen:
I realised that the issue with file size happened because I had tried to make the PR on raredisease test-data directly through GitHub. It works through terminal, so I will just update raredisease test-data.

PR #526, opened 2022-05-05
feat added fasta ref with chr20 and chrM

Lucpen:
Adding chrM to chr20 to run mitochondrial workflow


jemten (commented):
Nice work @Lucpen.
I know that it might be a little bit out of scope of your PR but I think some of the mitochondrial files that are currently in the `testdata` folder should be moved to the `reference` folder. Could you consider moving them in this PR or open up an new one for that`

jemten (approved):
🎉

PR #530, opened 2022-05-06
Adding an SDF of genome 21 (homo_sapiens)

nvnieuwk:


nvnieuwk:
@FriederikeHanssen Is it okay to merge?

PR #538, opened 2022-05-13
Added antitarget regions in multi_intervals.bed

SusiJo:
Added more antitarget regions to the existing `multi_intervals.bed` file. 
This would be important to for testing the cnvkit modules properly and helps avoiding empty antitarget files.


FriederikeHanssen:
🚀 awesome. Thank you 🙏

PR #547, opened 2022-05-24
adding working cellranger test-files

fmalmeida:
Hi guys,

I am wanting to add a couple of renamed files and organization of test-datasets that are working with cellranger so we can actually have a `-profile test` using it.


fmalmeida:
It is possible that these files also work with the others. If we check it later, we can then remove duplicity and have a single samplesheet for all.

PR #563, opened 2022-06-20
module vcf2maf: Added working VEP cache with no annotations

adamrtalbot:
Adds working minimal VEP cache for using with vcf2maf module (and possibly VEP module in future).

By including these two files, VEP runs correctly but doesn't annotate. I've reduced it to just chr22 to match the available test VCF and FASTA.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ If you cannot find suitable test data on this repository, please contact us on t
 ## Data Description
 
 ### genomics
+
 - sarscov2
 
   - alignment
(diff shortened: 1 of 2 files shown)


FriederikeHanssen:
Can you add it to the two Readmes: 
* Here for the general overview https://github.com/nf-core/test-datasets/blob/modules/README.md
* Here for how the data was generated https://github.com/nf-core/test-datasets/blob/modules/data/genomics/homo_sapiens/README.md (in case something needs to be adapted later it makes everyones life easier :) )

I guess no way of somehow basing this on the existing fasta?

FriederikeHanssen:
perfect 🎉

FriederikeHanssen (approved):

FriederikeHanssen:
woops merged already 🤠 it'll be fine

PR #570, opened 2022-07-14
Added interleaved FASTQ file

adamrtalbot:
Add an interleaved FASTQ file from the existing SARS-nCOV2 Illumina FASTQ data.


FriederikeHanssen:
can you add it to the readme?

PR #578, opened 2022-07-25
Add test files for hAMRonization/ABRicate

jasmezz:
Adding test file for hAMRonization/ABRicate which was succuessfully tested locally. The file is derived from running ABRicate on the _B. fragilis_ genome: `abricate genome.fna.gz > genome.abricate.tsv`
This output is the new test data for hAMRonization/abricate.


jfy133:
This would not go under the `genome` folder as it's not a genome related file. You could consider putting it in a folder called `tsv` instead.

jfy133 (changes requested):
see comment above

jfy133:
Wait, there is already an existing module: https://nf-co.re/modules

jfy133 (approved):

PR #579, opened 2022-07-29
Added a bgzip-compressed version of genome.fasta

muffato:
This will be used for testing software that require a bgzip-compressed file, e.g. `samtools`, c.f. https://github.com/nf-core/modules/pull/1935 (which is not happy with the existing gzip-compressed files)


jfy133 (changes requested):
Please update README describing what the file is

muffato:
Ah, yes, of course ! Sorry I forgot about that

jfy133 (approved):

PR #590, opened 2022-08-16
test file for goat/taxonsearch module

alxndrdiaz:
This .txt file contains the following species names and can be used as input for [goat/taxonsearch module, PR #1866](https://github.com/nf-core/modules/pull/1866): 

```
Saccharomyces cerevisiae
Arabidopsis thaliana
Caenorhabditis elegans
Drosophila melanogaster
Mus musculus
```


jfy133 (changes requested):
This is a biological file rather than purely a generic file, so I would put this back in `delete_me`. 

If it works, you could in fact also maybe put this file with the offical NCBI taxonomy name for SARS-Cov2 in it, in the `sars-voc2` directory somewhere to make it more reusable by other tools (as this file I could easily imagine being reused)

nvnieuwk:
> This is a biological file rather than purely a generic file, so I would put this back in `delete_me`.
> 
> If it works, you could in fact also maybe put this file with the offical NCBI taxonomy name for SARS-Cov2 in it, in the `sars-voc2` directory somewhere to make it more reusable by other tools (as this file I could easily imagine being reused)

I agree with this :)

alxndrdiaz:
> This is a biological file rather than purely a generic file, so I would put this back in `delete_me`.
> 
> If it works, you could in fact also maybe put this file with the offical NCBI taxonomy name for SARS-Cov2 in it, in the `sars-voc2` directory somewhere to make it more reusable by other tools (as this file I could easily imagine being reused)

Done, moved the file to `data/delete_me/goat_taxonsearch/taxonomy_ids.txt`

jfy133:
Ok, on the otherhand, I just remembered that we basically already have the same thing here: https://github.com/nf-core/test-datasets/blob/modules/data/genomics/sarscov2/genome/db/maltextract/taxon_list.txt

Why don't we consolidate these? We could replace `maltextract` with `NCBI` and extend the taxon_list.txt to have the contents of your current file? 

How does that sound?

alxndrdiaz:
> Ok, on the otherhand, I just remembered that we basically already have the same thing here: https://github.com/nf-core/test-datasets/blob/modules/data/genomics/sarscov2/genome/db/maltextract/taxon_list.txt
> 
> Why don't we consolidate these? We could replace `maltextract` with `NCBI` and extend the taxon_list.txt to have the contents of your current file?
> 
> How does that sound?

I think it would be better to keep it as a separate file, considering that the module (goat/taxonsearch) uses [goat-cli](https://github.com/genomehubs/goat-cli) that can only query data from eukaryotic taxa in [Genomes on a Tree (GoaT)](https://goat.genomehubs.org/) database.

jfy133:
Ok fair enough! You can merge

jfy133 (approved):

PR #599, opened 2022-09-06
feat: add minimal kraken2/Bracken database

Midnighter:
I suggest to add a kraken2/Bracken database. This one comes from the modules test data and is based on the SARS-CoV-2 genome. So it's probably not well suited to taxprofiler but is hopefully enough for the tests.


jfy133 (changes requested):
If possible please document it on the main README (so we know how to reconstruct it if we ever need to ;) )

Midnighter:
I really wish we would use pipelines to create these test data, too. Always so painful to set up.

jfy133:
> I really wish we would use pipelines to create these test data, too. Always so painful to set up.

Maybe we should add a db creation workflow... would only be another 8 modules :grimacing:

jfy133 (approved):

PR #607, opened 2022-09-09
Modules

ashotmarg:
This is an input "sites" file for testing the somalier/extract module. James, @jfy133, please have a look.


jfy133 (approved):

jfy133 (approved):

ashotmarg:
Hi @jfy133, the Somalier package is already in Bioconda 👍  Which means I am ready to update the Somalier Extract module.  However, we still need to figure out what to do with the required input file "sites" for testing. At the moment I can test it linking it to my own "test-datasets:modules/data/delete_me/": https://github.com/ashotmarg/test-datasets/blob/modules/data/delete_me/somalier/sites_chr21.hg38.vcf.gz
Or should I wait until my delete_me branch is merged with the nf-core's ?

jfy133:
As I've approved it, you can already merge this PR in and use the offiical nf-core URL :)

(let me know if you don't have merge permissiosns)

jfy133:
> I don't think I have merge/write permissions, can you give me those :)

Ok, will look into that now but I will merge in the meantime

> Also, I think I was a bit impatient in terms of the Bioconda for the Somalier. It seems that there is only a "placeholder" at the moment for Somalier in Bioconda... Hopefully soon it will be active, and this would cover the "conda" and "docker" for the containers. How about "singularity"? What should I do for this? I think that when Somalier is ready at quay.io/biocontainers/somalier:, it should automatically appear in "https://depot.galaxyproject.org/singularity/", right?

Yes, I think I posted this as a comment on the somalier/extract module. 

Bioconda exists (so you can still prepare the whole module etc, e.g. getting the test-yml etc. if you haven't already), but may have to wait a couple of days for the biocontainer (official docker/singularity) to appear. I sometimes in the past had to nudge poeple ;).

But yes, once docker appears, singularity should appear  in `dept.galaxy.org/singularity`.

ashotmarg:
@jfy133, thanks a lot for the revision, I will wait few days and as soon as the docker/singularity is available, I will get in touch!

PR #608, opened 2022-09-10
fix #603 including svd index stack into main test data

### LINKED ISSUE #603: Including indexing files for new module verifybamid
Including premixed SVD index and Generated SVD index for new module testing. 

nf-core/modules#1957

manifestoso:
including svd index derived from test.genome.vcf into main test dataset. fix #603 in prepare for nf-core/modules#1957


jfy133 (commented):
Folder structure looks OK to me 👍. However is the `genome` necessary in the file names? If not, I would remove that to make the names simpler.

You also need to update the README to describe what the files are (e.g. how generated).

jfy133 (approved):

PR #612, opened 2022-09-13
Update the path to remote

rafalstepien:
As we talked with @jfy133 , this path probably needs to be remote


jfy133 (approved):

rafalstepien:
@jfy133 can you also merge this? I don't have the permissions :grin:

PR #629, opened 2022-10-04
added constant region reference

ssnn-airr:
I have updated the reference sequences to include the constant region. Needed for IgBLAST >= 1.18.0.


ggabernet (approved):
Looks good to me!

PR #637, opened 2022-10-07
Expression matrix

pinin4fjords:
Adding a test dataset for downstream gene expression analysis, per discussion at https://nfcore.slack.com/archives/CJRH30T6V/p1665153847518509. 

Process of derivation described in a README, but I wasn't quite sure of the directory structure I was aiming for, so that might need revision.


drpatelh (approved):

drpatelh:
Was going to suggest having a more generic folder name like `counts` instead of `rnaseq_expression` but it's all good.

drpatelh:
Nice work with the docs too 🙏🏽

PR #638, opened 2022-10-10
Add new file for HaploCart and update README

JoshuaDanielRubin:
Test input file for HaploCart. The input file is simulated single-end FASTQ from the rCRS reference genome.


jfy133 (approved):
You don't need to say specifically for haplocart, and it might be nice to mention the tool(s) you used for generating the synthetic reads but otherwise LGTM!

PR #648, opened 2022-10-11
Fix airrseq test data

ggabernet:


snafees:
Thanks for adding the non-twice-gzipped files, woohoo!

snafees (approved):
Looks good!

PR #652, opened 2022-10-11
Adding support files for kofamscan, restricted to SARS-COV-2

toniher:
Data files needed for a minimum test for supporting https://github.com/nf-core/modules/issues/2189
Data processed from https://www.genome.jp/ftp/db/kofam/ restricting SARS-COV-2 proteome fasta hits.


toniher:
I was wondering to provide this in gzipped and tar.gz format respectively instead for reducing the number of files, even though the program would not accept it straight. Do you have some recommendations in that sense? Should then the module consider gzipped files to be processed in advance? Thanks!

toniher:
Thanks. I would use then both ```GUNZIP``` and ```UNTAR``` in the test but not in the module...

JoseEspinosa (approved):
LGTM 🚀

PR #660, opened 2022-10-14
Added test data for comradesnf

JLP-BioInf:


JLP-BioInf:
Hello, 

I am juts testing a few things at the moment but I am narrowing down on the bets place to host this pipeline. I think it should end up with nf-core but it would be great if I could test out a few things first. 

Al the very best, 

Jon - jlp76@cam.ac.uk

PR #664, opened 2022-10-17
Add input bam files for nanomonsv.

awgymer:
I am hoping to add new modules to nf-core for `nanomonsv`
This is an initial commit of test input bams for the `nanomonsv parse` command. 
The data used comes from the [original source code repo](https://github.com/friend1ws/nanomonsv/tree/master/tests/resource/bam). 
The output generated by the module matches on content the data in the source code repo but the md5 sums are different due to ordering changes (the source code tests just check for entries).


Emiller88:
Why'd you put these under `delete_me`, I don't know what the history of `delete_me` is anyway 😬

PR #668, opened 2022-10-19
refactor: ensure that kraken2 and bracken use the same DB

Midnighter:
Suggestion to change the databases sheet to ensure that kraken2 and Bracken **must** use the same database.


jfy133:
Conclusion: keep just `bracken` but must make sure database name is distinct from kraken2 ones! i.e. not just `db1`. We will add a schema check in taxprofiler for this

PR #669, opened 2022-10-20
Update contig names in ploidy priors

ryanjameskennedy:
The contig names column of the ploidy priors files were updated to be prefixed with 'chr'.


ramprasadn (approved):
LGTM

PR #670, opened 2022-10-20
Add info for testing blocking

pinin4fjords:
This PR adds functionality for testing blocking factors in differential analysis:

- A column added to the sample sheet to use as a blocking factor (in this case it might a sample pairing factor)
- A new row added to the contrasts with the blocking factor identified, to trigger a differential analysis with blocking factor included.
- Note added to README

I've also added 'id' column that we can propagate through to `meta.id`.


ggabernet (approved):
LGTM

pinin4fjords:
Thanks @ggabernet !

PR #677, opened 2022-10-25
added contig ploidy priors table

nvnieuwk:
Needed for gatk4/determinegermlinecontigploidy


jfy133 (approved):
Ok the homo_sapeins rEADME is new, but otherwise looks OK to me

nvnieuwk:
What do you mean with that? :)

jfy133:
I hadn't see that there was the per-species READMEs now (rather than the main README file).

PR #685, opened 2022-10-31
Create mini genes.gtf.gz for RNA-seq test outputs

pinin4fjords:
This PR adds a mini-GTF for use alongside the rna-seq test data.


maxulysse:
@pinin4fjords still needed?

pinin4fjords:
Probably not, since I'd forgotten about this- I'll close.

PR #689, opened 2022-11-11
Add spades test yaml input

TCLamnidis:
Adds a test yaml input for spades.


jfy133 (changes requested):
This is not a generic yaml file, it should go under `delete_me/spades`

jfy133 (approved):

PR #693, opened 2022-11-17
Openms idxml - README update

rolivella:
This is for updating README with the new data for this module. 

@JoseEspinosa let's see if now I did it right!!


JoseEspinosa (commented):
No, the PR should be done against the `modules` branch, not the master.
This time I did not change it so that you see it.
Also in principle, if you have committed to the same branch in your repository `openms_idxml` in your repository the commit should be added to the previous PR and not need to create a new one. Actually, I think the problem is that before you worked on the `proteomicsunitcrg` fork and this time in your own fork.

maxulysse:
No PR to master please

PR #703, opened 2022-11-29
UMI FASTQ file

adamrtalbot:
UMI FASTQ file ~composed of random 9bp synthetic oligos, all with uniform quality.~ 

Generated by stripping the UMI sequence from the existing FASTQ and turning it into a separate file. This will be a valid reference format for sequencing kits where the UMI is embedded in the index.


lescai (changes requested):
Hi :)
could you please describe a little more this file?
if this is the use case where UMIs are present in a third FASTQ, then the test dataset should include 3 files: forward and reverse (without UMIs in the sequence), and a UMIs file.

lescai:
also, UMIs structure is needed in order to process the sequences

maxulysse (approved):

famosab:
@lescai Would you be ok with the final version? 

If we get no reply I would suggest to open a new PR @adamrtalbot if the data is needed.

PR #712, opened 2022-12-09
Add test data for cellranger-arc (10X Multiome) module

dinhdiep:
Hello,

I create test data for the cellranger-arc pipeline which is used to analyze 10X Multiome data. 

Best,

Dinh


ggabernet:
Hi @dinhdiep , thanks a lot for the contribution! This PR seems to include a lot of commits. Could it be that you created your branch from the `master` branch instead of the `modules` branch first? These branches have separate history, so it would be great if you could create a new branch from the `modules` branch and add you changes directly there

PR #715, opened 2022-12-13
Remove read numbers from read names for methylated FASTQs

njspix:
The sarscov2/illumina/test_methylated fastq files had read names such as:
`asdfasdf_R1` and `asdfasdf_R2`, which confused downstream tools which depend on reads from the same template molecule having the same name. Removed _R[12]$ from the end of each read name and recompressed these files.


muffato:
Hi. These files are used in several modules:

- biscuit/align
- biscuit/biscuitblaster
- bismark/align
- bismark/coverage2cytosine
- bismark/report
- bismark/summary
- bwameth/align

I don't want to bring their tests unintentionally. Could you check them and make a PR (or hijack https://github.com/nf-core/modules/pull/2656) if necessary ?

njspix:
Hello @muffato, thanks so much for the review! I pulled down the `modules` repo, edited `tests/config/test-data.config` to point to my fork, and ran local tests (using Singularity) for all the modules you suggested. I updated the `biscuit` modules on https://github.com/nf-core/modules/pull/2656. All the others passed without hiccups, and I checked their `test.yml` files - they don't contain m5d hashes for the .bam files, so this makes sense.
I should note that `bwameth/align` didn't pass, but that was due to some weird Singularity mount error. I also checked the `test.yml` for `bwameth/align` and it also doesn't have hashes for .bam files, so should be no problem

muffato (approved):

PR #726, opened 2023-01-09
Differentialabundance match samples with matrix

pinin4fjords:
This PR fixes the test data for the differentialabundance workflow:

- Suffix removed from matrix columns such that they can be related to the sample sheet
- Variable in contrasts file fixed to actually match what's in the sample sheet


erikrikarddaniel (approved):
LGTM.

pinin4fjords:
Thanks @erikrikarddaniel !

PR #733, opened 2023-01-23
Add computed genotyped likelihood

LouisLeNezet:
Here I added the computed genotyped likelihood file for the individual to be imputed for Glimpse modules.
The index is also added.
The detail of the procedure is present in the branch imputation of test-datasets.


jfy133 (approved):

LouisLeNezet:
Thanks for the fast review !

PR #743, opened 2023-02-03
Add full-size assembly test data for metapep

skrakau:
Add full-size assembly test data for metapep. For the beginning using only two samples for the full-size test (specified with the `samplesheet.fullsize_assembly.csv` file.


tillenglert (approved):
Looks good to me! Looking forward to benchmark with these!

PR #746, opened 2023-02-06
Restored featurecounts output format of counts table

WackerO:
This PR restores the original version of the counts table as it is produced by featurecounts (sorry I pushed directly to nf-core/test-datasets earlier instead of to my fork, that was definitely an accident)


pinin4fjords (changes requested):
This is a reversal of https://github.com/nf-core/test-datasets/pull/726, that PR has the reasons I had for that.

If you do this, we'll need to replace the #differentialabundance test data, which expects a matrix with column names that relate to the sample sheet.

pinin4fjords:
I see now that this is in light of https://github.com/nf-core/differentialabundance/pull/62, but I think we can accomplish those goals more simply.

WackerO:
Closing this as not necessary

PR #749, opened 2023-02-07
Add extra kraken2 file with different name

sofstam:
For `taxpasta/merge` module, we need at least two files from a classifier. It was suggested to add another kraken2 report file with different name and same contents.


jfy133 (changes requested):
Needs to be documented on README

Midnighter:
Should we close this then?

PR #751, opened 2023-02-07
Adding nanomonsv parse data for testing nanomonsv get

awgymer:
In order to test new module `nanomonsv/get` [modules issue](https://github.com/nf-core/modules/issues/2848) I will need these `parse` intermediates.


maxulysse:
Can you add details for the data in the README?

adamrtalbot (commented):
- Were these files generated from the existing BAMs using nanomonsv/parse? Is it possible to generate them during tests (not ideal but could save on space).
- Add the data to the data description, clarify how you made them.

awgymer:
> * Were these files generated from the existing BAMs using nanomonsv/parse? Is it possible to generate them during tests (not ideal but could save on space).
> 
>     * Add the data to the data description, clarify how you made them.

The data come from the `nanomonsv` repo itself. It's just the test-set the authors used themselves. Generating them on the fly would create a dependency on the `nanomonsv/parse` module so I'm not sure we want that. 

Where does `delete_me` data get documented - unclear where exactly to put the data description?

awgymer:
Closing per [slack thread](https://nfcore.slack.com/archives/CQY2U5QU9/p1675784070942519)

PR #756, opened 2023-02-09
fixed PRG_test archive

mapo9:
Something went wrong with the zipping the first time which resulted in an empty archive, fixed the problem and now reuploading the zipped input folder to create a PRG test graph


ggabernet (approved):
Thanks, looks good to me!

PR #759, opened 2023-02-14
Metapep expand coassembly test

tillenglert:
This PR extends the coassembly test to include also bins, for reduced runtime another allel was chosen for the assembly to enable the use syfpeithi.


testdata/test_coassembly.bin_weights_cond_2.tsv
```diff
@@ -0,0 +1,4 @@
+bin_basename	weight
+bin.1	0
+bin.2	5.21532
+
```
skrakau, line 4:
```suggestion
```

skrakau (approved):
LGTM

PR #760, opened 2023-02-16
update test dataset for raredisease 

ramprasadn:
Minimal test data to test all the functionalities of the  raredisease pipeline.


ramprasadn:
@maxulysse and @jemten When we discussed about this dataset in our last meeting, I had mentioned that total size of all files was roughly about 2.5GB. I have managed to bring it down to approx 500MB now 😄

maxulysse (approved):
Good job

PR #767, opened 2023-02-20
Change fileformat of taxa input to tsv, add weights to test tiny

tillenglert:
This PR addresses the taxa input and corresponding tests:

- Changes all file formats of the taxa input to tsv, as the input check was changed for this input type.

- Added new taxa input files, containing weights for test_tiny, to include variable weights for the same taxid.


skrakau (approved):
LGTM!

PR #769, opened 2023-02-20
Add TBI of SV VCF

nvnieuwk:


jfy133 (approved):
needs descripton added to the README but then GTG (have to run now, but I trust you to update it before merging)

nvnieuwk:
Oh I always forget those 😁. I'm on it!!!

PR #770, opened 2023-02-20
remove parents from single sample test

jemten:
The single sample test samplesheet had the parents still in. This PR removes them


ramprasadn (approved):
👍🏻

PR #775, opened 2023-02-24
Add descriptions to test fasta

jtangrot:
Added random descriptions to some sequences in test fasta file, one including a "tab" to test that this does not cause troubles in the results.


erikrikarddaniel (approved):
LGTM.

PR #777, opened 2023-02-24
initial commit adding the expanded reference files

peterpru:
These update reference fasta and target_wgs.interval_list contain a 1000bp sequence of NNN for all numbered chromosomes.


ramprasadn (approved):
LGTM

PR #778, opened 2023-02-24
Added an empty VCF

nvnieuwk:


README.md
```diff
@@ -329,6 +329,8 @@ If you cannot find suitable test data on this repository, please contact us on t
       - test.rnaseq.vcf: RNAseq vcf corresponding to `test.rnaseq_{1,2}` reads
       - test.genome_21.somatic_sv.vcf: Indels VCF corresponding to `test.paired_end.recalibrated.sorted` and `genome_21.fasta` generated with Manta
       - NA12878*chrM.vcf.gz: mitochondrial variants corresponding to `testdata/NA12878_mito*{1,2}.fq.gz`from the`rarediseases` branch.
+      - empty.vcf.gz: The RNAseq VCF with all variants removed
+      - empty.vcf.gz.tbi: The index of the empty vcf
```
adamrtalbot, line 333: The VCF header says it's a germline VCF (HaplotypeCaller).

adamrtalbot (approved):
I downloaded and checked the file and it looks good.

README.md
nvnieuwk: Strange, I picked the vcf that was called rnaseg.vcf.gz

adamrtalbot: It doesn't matter for most things but sometimes the FORMAT or INFO fields can be different.

nvnieuwk: Yeah since there are no variants in here it shouldn't matter indeed

PR #783, opened 2023-03-07
Add reference genome from needed region for imputation

LouisLeNezet:
Hi,

Here is a small addition to the Glimpse delete_me folder for further testing with the GLIMPSE2_software.

Thanks


sofstam (approved):
Lgtm!

PR #789, opened 2023-03-13
Revert human_design_bam.tsv. Add human_design_bam_eager3.tsv

TCLamnidis:


jfy133 (changes requested):
Call it eager3 rather than DSL2 (the samplesheet is not specific to DSL3)

PR #806, opened 2023-03-15
update paths taxprofiler

FriederikeHanssen:


ewels (approved):
@jfy133 - the switch from https:// linkst to s3:// links should be fine, right? Just wanted to flag in case stuff breaks..

jfy133:
> @jfy133 - the switch from https:// linkst to s3:// links should be fine, right? Just wanted to flag in case stuff breaks..

Yeah probably fine. Mistake from me in the past!

PR #809, opened 2023-03-15
update aws paths airrflow

FriederikeHanssen:


ggabernet (changes requested):
Hi @FriederikeHanssen, thanks a lot for taking care of this. Unfortunately it seems like a space was introduced before each s3 path, otherwise looks good to me!

FriederikeHanssen:
good catch 🙈 , fixed

ewels (approved):

PR #828, opened 2023-03-29
Add test directory for kmcp module

sofstam:


jfy133 (changes requested):
The uploaded file should be a `tar.gz` archive, that contains a directory that has the two files.

You can't stage 'directories' from URLs

sofstam:
I updated the PR, hope this is what you meant :)

jfy133:
Yes!

jfy133 (approved):

PR #833, opened 2023-03-29
Fix Background Subtraction markers file

adamjtaylor:
markers.csv for Background Subtraction testing had 12 channels while the image had 8
This PR removes the last 4 channels  (third cycle from the markers.csv


kbestak (approved):
Looks good!

PR #835, opened 2023-03-29
Create minimal samplesheet with updated schema 

emmarogge:
Create samplesheet with updated schema ('sample' column instead of 'group' and 'replicate' columns) for minimal (yeast) dataset which works with recent versions of pipeline.


samplesheet/samplesheet_minimal.csv
```diff
@@ -0,0 +1,9 @@
+sample,fastq_1,fastq_2,strandedness
+	sample
```
maxulysse, line 2: Something is wrong there

maxulysse (changes requested):
I think there's an issue in the samplesheet

PR #843, opened 2023-03-31
Add a (pseudo) NGSCheckMate bed file

SPPearce:
Adds a tiny bed file that contains 5 positions covered by the test dataset, for the use of testing NGSCheckMate. 
If someone happens to have an idea on some common SNPs in yeast then that could be used instead, but don't think that is required.


klkeys (approved):
looks ok so far

not sure if you are still working on this @SPPearce but just in case I'll let you merge at your leisure

SPPearce:
Thanks! Getting this tool into rnaseq and sarek is on my to-do list

PR #851, opened 2023-04-13
Add gatk gcnvcaller test data

ryanjameskennedy:
Four tar gzipped directories required for GATK's module testing (DetermineGermlineContigPloidy, GermlineCNVCaller & PostprocessGermlineCNVCalls). File sizes are as follows:
```
64K     ./genome.germline_cnv_calls.tar.gz
48K     ./genome.germline_cnv_model.tar.gz
64K     ./genome.ploidy_calls.tar.gz
4,0K    ./genome.ploidy_model.tar.gz
```


ryanjameskennedy:
Indeed. Well it was initially our model that our group created using complete human genomes and not the test fasta (which is like 40kb). Also it was zipped (not tar gzipped).

maxulysse (approved):

PR #866, opened 2023-04-26
add datasets for ultraplex demultiplexer nf-core module testing

CharlotteAnne:
we use ultraplex demultiplexer for clip data demultiplexing, clip data contains 5' and 3' barcoding with UMIs that must be moved to read header, therefore test data already in the demultiplex branch was not suitable. the demultiplexing also requires a barcode file which i've added.


CharlotteAnne:
merging for now to move on with pipeline dev - if anyone has an issue please let me know and i can move the data.

PR #870, opened 2023-05-04
Update MAG full samplesheet

jfy133:


d4straub (approved):
Are you sure of 
`s3://ngi-igenomes/test-data/mag/input_data/ERR3201918_1.fastq.gz`
because in https://github.com/nf-core/mag/issues/415 is using instead
`s3://ngi-igenomes/test-data/mag/ERR3201918_1.fastq.gz`
(I have no idea so I approve anyway)

PR #887, opened 2023-05-15
Add a GSE ID (GEO) to sra_ids_test.csv

joshiesh:
Adds a small GEO ID [`GSE214215` ](https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSE214215)for a series/experiment that is composed of 2 SRA runs. Allows us to test both `GSM*` and `GSE*` IDs as they are resolved differently.

- Addresses this issue https://github.com/nf-core/fetchngs/issues/104 
- Relevant PR here https://github.com/nf-core/fetchngs/pull/155 which adds back functionality for using GEO IDs in this pipeline


drpatelh (approved):
🚀

PR #895, opened 2023-05-22
Cellranger-Arc test data for cellranger-arc modules

heylf:
I took data from this PR (https://github.com/nf-core/test-datasets/pull/713) and added things for the needs for the cellranger-arc modules. The modules work with this test data.

I did not created this data and only renamed it and added a required lib.csv.


heylf:
I need to check something and I will reopen after my holiday.

PR #899, opened 2023-05-26
Add nodes.dmp and names.dmp to folder

sofstam:
One of the input arguments of `kmcp profile` is:
 
```
 -X, --taxdump string      Directory of NCBI taxonomy dump files: names.dmp, nodes.dmp, optional with merged.dmp and delnodes.dmp.
```

This PR adds a folder in zipped format including `names.dmp` and `nodes.dmp`


jfy133:
Maybe this should go into a kmcp folder?

sofstam:
I thought that the folders should be always zipped but I can change to a kmcp folder instead.

jfy133 (approved):

jfy133 (changes requested):
Ah sorry, missed your question: no you're right, if you need a folder, then yes a `tar.gz` is needed.

However what I meant is that you should have `delete_me/kmcp/kmcp.tar.gz` for example, (rather than `delete_me/kmcp.tar.gz`)

sofstam:
Ok, I will update the PR.

jfy133 (approved):

PR #901, opened 2023-05-29
Airrflow

ssnn-airr:
Removed mouse test. Renamed from reveal to assembled, since the reveal subworkflow has been replaced with --mode assembled


ggabernet:
@ssnn-airr Sorry I forgot to say this yesterday, but we should not change the filenames or directory names of the test repository as this breaks the pipeline tests for the previous releases. So it is better to add a new samplesheet with the removed mouse samples, and give it a different name and specify only the mouse/human samples there.

ssnn-airr:
I went back to the previous naming (reveal) and added two new samplesheets, one for human and one for mouse.

ggabernet (approved):
LGTM

PR #904, opened 2023-05-30
Correct scientific name in metagenome/names.dmp

sofstam:
This PR corrects the scientific names in `names.dmp` so that it matches the `seqid2taxid.map`


jfy133 (approved):
I'm tentaively approving this as it makes sense, but I vaguely remember thsi might break things somehwere else (I remember something like Ihad to point 0.1)... but lets see!

sofstam:
Or I can leave this as it is and just change the scientific name in `names.dmp` in `delete_me` folder.

PR #919, opened 2023-06-05
Add Illumina format SampleSheet to repository

adamrtalbot:
- Adds Illumina format SampleSheet for using with `-test profile` which currently doesn't work because of relative paths. See https://github.com/nf-core/demultiplex/issues/103


matthdsm:
Isn't this file already defined in `https://raw.githubusercontent.com/nf-core/test-datasets/demultiplex/testdata/MiSeq/SampleSheet.csv`?

PR #920, opened 2023-06-05
Move nf-core/demultiple test data from pipeline to test-datasets

drpatelh:
Closes https://github.com/nf-core/demultiplex/issues/103

### DIFF
--- a/samplesheet/1.3.0/samplesheet_full.csv
+++ b/samplesheet/1.3.0/samplesheet_full.csv
@@ -1,0 +1,2 @@
+id,samplesheet,lane,flowcell
+35147139,https://raw.githubusercontent.com/nf-core/test-datasets/demultiplex/testdata/miseq_35147139/miseq_35147139_samplesheet.csv,1,s3://ngi-igenomes/test-data/demultiplex/miseq_35147139.tar.gz


maxulysse (approved):

maxulysse:
Is the samplesheet format unchanged since 1.0.0 ?

drpatelh:
We add a new folder for 1.4.0 if required. If not, we carry on using the 1.3.0 test data until we have to change it. This at least shows you the last version in which the test data was changed.

drpatelh:
> I think that's misleading - the data version number will be different from the pipeline version number.

Don't know of a more intelligent way to track test-data updates without making in-place changes that will break CI for older releases. Agree it's not perfect but it does the job, especially since test data isn't updated very frequently. Open to suggestions.

drpatelh:
Also, at the moment, the samplesheets for this pipeline were all over the place:
- Within the pipeline
- Within this repo in the top-level folder as well as within `samplesheets/`

This unifies where to find all of the samplesheets in a single location.

maxulysse:
> > Is the samplesheet format unchanged since 1.0.0 ?
> 
> Yes, they look unchanged according to commit history.

Then let's put in a `1.0.0` folder, that way, we know it's csv that works since `1.0.0`

drpatelh:
> Yeah it's definitely an improvement. I would probably just go v1, v2, v3 etc. completely independently from the pipeline version but that's not very intuitive either.

Exactly. This way we know in which release it was last updated.

> Then let's put in a 1.0.0 folder, that way, we know it's csv that works since 1.0.0

Nope, there are other samplesheets in this new folder that were changed/added more recently in later releases. Let's start with a blank slate with `1.3.0` to indicate when everything was collated together.

PR #921, opened 2023-06-06
Adding SB26 full dataset

matq007:
In this PR I'm adding big boy test dataset. The old SB26 (which was subset) was removed accordingly for less confusion.


FriederikeHanssen (commented):
can the sb26/.xlxs files be removed again snce they are on aws?

matq007:
> can the sb26/.xlxs files be removed again snce they are on aws?

I was thinking maybe it makes more sense just to keep the fastq files on AWS as Phil mentioned. What do you think? I have a helper script which parses the excel files, but if they are on `s3` I found out I have to install one additional library to read them using `pandas`.

FriederikeHanssen:
sounds good to me. HOwever how would it work then for people that run everything on AWS later 'in production'?

FriederikeHanssen:
I mean for a real case, will it work on AWS alone? Because if people process everything on AWS they will have their excelsheets in an s3 bucket

matq007:
Ok, so it doesn't work when the excel files on aws. I would have to create a new mulled container with missing library. For the sake of simplicity, can we just discard the excel files from AWS, and people just use the excel files from the test-dataset?

FriederikeHanssen:
Ok for the test data now it is likely ok-ish but for the final release it really should work.

PR #922, opened 2023-06-07
Add technical replicate for spaceranger example data

grst:
Useful for testing if specifying multiple fastqs for a single sample works.


klkeys (approved):
looks good

PR #931, opened 2023-06-15
Add dumpfiles to create metabuli database

TheOafidian:


jfy133 (commented):
Seems a bit wierd to have all the `nan` and true/false, but I'll trust you :grimacing:

TheOafidian:
Ah yes, I did the filtering of the taxdump files with scikit-bio, but there it does seem to read it into a numpy array and output empty fields therefore as nan and the 0-1 flags as True/False... I've changed these values to fit with the expected format, thanks!

jfy133 (approved):

PR #942, opened 2023-06-21
Add gene score test YAML

apeltzer:
This adds an example test for our gene score functionality in Nanostring


apeltzer:
Required for https://github.com/nf-core/nanostring/pull/32

christopher-mohr (approved):
LGTM, we could put it in a sub folder but if we do not plan to add more it is not necessary.

apeltzer:
No, one is enough I think 👍🏻

PR #943, opened 2023-06-21
Add archived minimal database

TheOafidian:
As the metabuli database building takes up too much memory for CI purposes, I propose to add a prebuilt minimal db (of just the sarscov entry) to the test-datasets, so that the classify module can be tested, while the database build module can just run as a stub.


jfy133 (approved):
We are starting to request READMEs along files describing what is in such files and how they were generated (basically what genome, and what command - right @FriederikeHanssen ?)

Could you please add that (maybe Rike can add an example)? Otherwise L:GTM :+1:

jfy133:
Also @FriederikeHanssen can you bring this up next maintainers meeting, to come up with a standard tempalte and guidelines on it?

PR #944, opened 2023-06-22
upload goldset/truth file and respective query file for somatic analysis

LlaneroHiboreo:


maxulysse:
For which modules is this used?
What motivated the changes?

PR #945, opened 2023-06-22
Test dataset with mitochndrial data

ksenia-krasheninnikova:
Dataset contains mitochondrial contigs and raw HiFi reads of _Deilephila porcellus_ , as well as the reference dataset MW539688.1. Will be used for testing MitoHiFi.


priyanka-surana (changes requested):
Add a description in this README too: https://github.com/nf-core/test-datasets/blob/modules/README.md

priyanka-surana (approved):
LGTM 👍🏼

PR #952, opened 2023-06-29
Add taxdump files seperately

sofstam:
All those three files need to be provided in `test.config` and thus is better to have them separately in the folder and not in a zipped folder.


sofstam:
In `nextflow.config`, we need to give the path to taxdump files (names.dmp, nodes.dmp) in order for kmcp to be executed. 
For example:
```
   // kmcp
    run_kmcp     = false
    kmcp_taxdump = null
    kmcp_taxid   = null
    kmcp_mode    = 3      // default  kmcp profiling value
```

PR #958, opened 2023-07-18
Add compressed tabular text format files for seqkit tests

heuermh:


heuermh:
Thank you, @adamrtalbot

PR #959, opened 2023-07-19
Rebuild compressed db

TheOafidian:
Rebuild the compressed db from the command line according to the Readme instructions given.


FriederikeHanssen:
You are rebuilding, but no file is replaced, is that right?

PR #962, opened 2023-07-20
Add test data for BAM input

fa2k:
Data for pipeline test with bam. See: nf-core/raredisease#261


jemten:
sorry @fa2k, I just noticed this PR. Should we close it since the corresponding PR in nf-core/raredisease is closed?

fa2k:
Yes I also missed this one. I'll close it because I think the sample sheet format is likely to be different than this one, so it's better to make a new one when the bam input comes around again.

PR #968, opened 2023-07-27
Test data for scnanoseq

atrull314:
Hi,

This is the commit for the minimal test dataset for the scnanoseq pipeline that was approved and in development. The pipeline is currently being validated, but wanted to get the ball rolling on some of the other required steps for pipelines.

Let me know if there are any questions

Thanks,
Austyn

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -13,8 +13,8 @@ authors:
     given-names: Johannes
   - family-names: Wilm
     given-names: Andreas
-  - family-names: Ulysse Garcia
-    given-names: Maxime
+  - family-names: Garcia
+    given-names: Maxime Ulysse
   - family-names: Di Tommaso
     given-names: Paolo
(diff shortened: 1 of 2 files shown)


asp8200:
Looks okay to me, but apparently there are som conflicts that needs resolving

SPPearce (commented):
Why are there a bunch of files being deleted from the testdata folder? Where did they come from, is anyone using them in a module?

atrull314:
> Why are there a bunch of files being deleted from the testdata folder? Where did they come from, is anyone using them in a module?

It looks like its due to the timing of when we forked the repo. Testdata had been added in master and was removed after we had forked. Its not there anymore and we don't need the data that was/is there for the pipeline

SPPearce (approved):
Sorry, I did get the request notification last week but not at a point I was free. And then completely forgot about it.
Looks good to me.

PR #973, opened 2023-07-28
eager: Add new nf-validation compatible samplesheet

jfy133:
Add samplesheet that has empty strings instead of `NA`


TCLamnidis (approved):
LGTM

PR #982, opened 2023-08-03
Ci tests cleanup

tillenglert:
I changed some input sample sheets according to the new tests on nf-core/metapep (https://github.com/nf-core/metapep/pull/93)


skrakau:
It's a bit confusing, there is a `input.bins.csv` and `input.bins_only.csv` now. I guess in the past `bins` was not tested in the main test profile and called `input.bins.csv`, but since `bins` are also tested in the main test profile, this is probably not necessary anymore, right? So one could just delete one

tillenglert:
These files are not used in any test:

- https://github.com/nf-core/test-datasets/blob/metapep/testdata/fullsize_CAMISIM_SPAdes.G0_T2.top300000.fasta.gz
- https://github.com/nf-core/test-datasets/blob/metapep/testdata/fullsize_CAMISIM_SPAdes.G0_T2_depths.top300000.tsv

Should I delete them? We still have T0 and T1!

skrakau:
> These files are not used in any test:
> 
> * https://github.com/nf-core/test-datasets/blob/metapep/testdata/fullsize_CAMISIM_SPAdes.G0_T2.top300000.fasta.gz
> * https://github.com/nf-core/test-datasets/blob/metapep/testdata/fullsize_CAMISIM_SPAdes.G0_T2_depths.top300000.tsv
> 
> Should I delete them? We still have T0 and T1!

Yes, do it :)

skrakau:
Since the samplesheet for the full-size test ist called `samplesheet.fullsize_assembly.csv`, would it maybe make sense to call all those files `samplesheet.XX.csv` instead of `input*`?

skrakau:
Why exactly is `input.mouse_assembly.csv` needed?

skrakau:
Not sure, could one maybe improve the naming scheme in `test-datasets/testdata`?

tillenglert:
I tried cleaning up the files again, and renamed all input. ... to samplesheet. ...! Maybe you can shortly look at it again @skrakau?

skrakau (approved):
LGTM!

PR #987, opened 2023-08-16
add data to test quilt module

atrigila:
Added data for testing the quilt module, which is currently under construction, and is likely to be also valuable for testing within the #phaseimpute pipeline.

### DIFF
--- a/data/delete_me/quilt/bamlist.1.0.txt
+++ b/data/delete_me/quilt/bamlist.1.0.txt
@@ -1,3 +1,3 @@
-package_2021_01_15A/NA12878.haplotagged.1.0.bam
-package_2021_01_15A/NA12878.ont.1.0.bam
-package_2021_01_15A/NA12878.illumina.1.0.bam
+NA12878.haplotagged.1.0.bam
+NA12878.ont.1.0.bam
+NA12878.illumina.1.0.bam


LouisLeNezet:
Hi @atrigila,

I have made a [imputation branch on the test-datasets repository](https://github.com/nf-core/test-datasets/tree/imputation).
Would you mind adding your scripts to generate those files in there ?

Louis

atrigila:
Hi @LouisLeNezet. The files are made available by QUILT for testing, but it was not shared how they were generated. They have the 1000 Genomes haplotypes and NA12878 bams. They are downloaded from http://www.stats.ox.ac.uk/~rdavies/QUILT_example_2021_01_15A.tgz. I am only adding them here (in modules) so that I can reproduce the same example as QUILT does, but I agree for the full pipeline test we can use your files available in the imputation branch.

LouisLeNezet (approved):
Let's go for them.
We will need afterwards to do some clean-up and update the test with the files of the pipeline.

saulpierotti:
What do you think about using the cram files at `https://github.com/nf-core/test-datasets/raw/modules/data/genomics/homo_sapiens/illumina/cram/test.paired_end.recalibrated.sorted.cram` and `https://github.com/nf-core/test-datasets/raw/modules/data/genomics/homo_sapiens/illumina/cram/test2.paired_end.recalibrated.sorted.cram`? To avoid creating more test files. I generated a posfile from `https://github.com/nf-core/test-datasets/raw/modules/data/genomics/homo_sapiens/genome/vcf/dbsnp_146.hg38.vcf.gz` by extracting the biallellic sites with bcftools query, I could upload it to the repo. And the samplefile could be just computed on the fly in the test pipeline with collectFile. Does quilt need additional inputs?

LouisLeNezet:
It would effectively be nice to use some of the nf-core available data.
I've face unfortunately multiple missing fields working with them (for the reference panel at least).
But in the end a big discussion will be needed :)

PR #1004, opened 2023-09-20
Replace ftp with https due to ENA ftp problems

jfy133:
FTP downloads with NXF is failing from ENA servers for some reason (idnetified during megatests and replciated locally).

Using HTTPS links are slower, but still worked - working locally.


LilyAnderssonLee (approved):
Great!

Issue #1050, opened 2023-11-29
Add new test data for modkit_pileup


Michal-Babins:
New data is needed for modkit_pileup module. 
The data needed is a bam file and a bai file.

PR #1051, opened 2023-11-29
Add test data for modkit/pileup module

Michal-Babins:
#1050 

Modkit/pileup output bed file was coming up empty with bam files available in the modules branch. The correct way to test is to use nanopore human test data. `bc_anchored_10_reads.sorted.bam` and `bc_anchored_10_reads.sorted.bam.bai` are provided by the `modkit/pileup` developers and creates a valid output while running quickly.


ggabernet (approved):
Looks good to me!

PR #1054, opened 2023-12-05
Add fastqc outputs for testing multiqc

pinin4fjords:
Just adding some fastqc outputs. Our use case is to allow testing of multiqc at the pipeline level without having to have the fastqc module installed.


FriederikeHanssen:
docs please :)

pinin4fjords:
Thank you @FriederikeHanssen :-)

PR #1056, opened 2023-12-08
Depricate variant tsv epitopeprediction

jonasscheid:
According to discussion in https://github.com/nf-core/epitopeprediction/pull/220


christopher-mohr (approved):
LGTM

PR #1058, opened 2023-12-15
Added full test samplesheet.

FloWuenne:
This PR adds the samplesheet for running two full sized samples from:

https://www.nature.com/articles/s41467-023-43005-8

available via Zenodo:

https://zenodo.org/records/8413573


kbestak:
LGTM

PR #1063, opened 2024-01-09
adding mobile element references

jemten:
Adding test data for mobile elements


peterpru (approved):
in the file reference.fasta.fai there is no chromosome X, that is correct?

jemten:
Very observant! I basically just ran samtools faidx on `reference.fasta` and since that file lacks chrX the index file also lacks it. It looks like you where the one adding the original file https://github.com/nf-core/test-datasets/pull/777 😉 
It should work but otherwise I'll make a new PR updating the reference

PR #1067, opened 2024-01-12
Add Illumina NovaSeq 6000 test runfolder

Aratz:
There is currently only a MiSeq run available for testing nf-core/demultiplex. However, the file structure of a runfolder differs slightly depending on the instrument it comes from. As such , we should aim at collecting test datasets from as many instruments as possible.

This PR adds a test dataset from a NovaSeq 6000 instrument that was downloaded from Illumina public repository. The `cbcl` files have been edited using [this script](https://github.com/Aratz/cbcltrimmer) so as to only include the first tile of the first lane in each cycle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,10 @@ This folder contains data from a MiSeq run. The data has been trimmed to only
 keep the first tile in each cycle. To demultiplex this data, you will need to
 use `--tiles s_1_1101` with `bcl2fasq` or `--first-tile-only true` with
 `bclconvert`.
+
+#### NovaSeq 6000
+
+This folder contains data from a NovaSeq 6000 run. The data has been trimmed to
+only keep the first tile in each cycle. To demultiplex this data, you will need
+to use `--tiles s_1_2101` with `bcl2fasq` or `--first-tile-only true` with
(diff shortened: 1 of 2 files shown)


maxulysse (approved):
LGTM

docs/ADD_NEW_DATA.md
edmundmiller, line 1: Maybe we should move these to PR Templates?

edmundmiller (approved):
Looks good to me!

docs/ADD_NEW_DATA.md
maxulysse: that's a good point, I don't that existed when we created this repo

Aratz: Seems like they've deleted these files on the methylseq and rnaseq branches. Maybe we could do the same?

maxulysse: We should do some "template" sync in between all the branches

Aratz: Alright, I'll merge this now then and create a separate issue for harmonizing these files. Thanks for the review!

PR #1075, opened 2024-01-31
add genmod score vcf file

ramprasadn:
Genmod module requires vcf records to contain certain annotations, and the file in this PR contains them.


jemten (approved):
👍

PR #1082, opened 2024-02-05
Add NovaSeq 6000 flowcell

Aratz:
Adds a dataset from a NovaSeq 6000 flowcell.

This data set will be used to test the bclconvert and bcl2fastq modules more thoroughly as the flowcell structure differs slightly between e.g. MiSeq and NovaSeq 6000. Specifically the new flowcell contains by-cycle Interop files.


matthdsm (approved):
Great addition!

maxulysse (approved):

Aratz:
Thanks @matthdsm @maxulysse !

PR #1085, opened 2024-02-06
eager: fixed RG tags in latest bam

TCLamnidis:
Fix for #1084 . Seems RG tags were only replaced in header and not in reads. This fixes that.


TCLamnidis:
@jfy133 ping

jfy133 (approved):

PR #1095, opened 2024-02-12
Added chr1 of actinidia_chinensis for LAI module

GallVp:
The current genome files are either too short or soft masked. The LAI module (https://github.com/nf-core/modules/pull/4771) works on the repeat space and requires an unmasked genome. After trying various increasingly smaller sizes: chr1 whole, 14 million bases, 7 million bases, 7 thousand bases, 7 hundred bases; I decided to use 7 million bases as this is the tallest size for which the LAI module produces sensible outputs.


mahesh-panchal:
According to the guidelines https://nf-co.re/docs/contributing/test_data_guidelines I think there should be no chr1 dir. Also test data doesn't need to produce sensible output on the module it's used with. The question is, if used with other datasets, does this produce a lot of empty files, or errors when used with the module?

GallVp:
Thank you @mahesh-panchal

Yes, the module does not produce any output with the old data. The key output file is empty as asserted by this test: https://github.com/nf-core/modules/blob/9e4b1450c047d6a3facde4cc2b9170f542f3c723/modules/nf-core/lai/tests/main.nf.test#L80

GallVp:
I included chr1 dir following the organisation of homo_sapiens data: https://github.com/nf-core/test-datasets/tree/modules/data/genomics/homo_sapiens/genome

Later on for some other module we might have to create a small version of the entire genome of *Actinidia chinensis*. Such a genome might include subsequences of all the chromosomes. Having a a chr1/genome.fasta.gz future proofs this scenario. What do you think? I'm happy to remove the chr1 dir.

mahesh-panchal (approved):
Let's go with that then.

PR #1099, opened 2024-02-15
CREATETAXDB: add nucleotide taxdump files

jfy133:


data/taxonomy/nucl2tax.map
sofstam, line 1: Shouldn't this file have the same content as names and nodes?

sofstam (approved):

data/taxonomy/nucl2tax.map
jfy133: AHH forgot the influenza one yes

jfy133: Oh no influenzae is there.. no doesn't need to have all names/nodes, this matches the GenBank/RefSEq accession to the taxonomy - so just one entry per fasta header acccesion.

PR #1100, opened 2024-02-16
single-cell readme added

mapo9:
After adding the single-cell dataset, the respective README part is added with this PR


ggabernet (approved):
LGTM

PR #1109, opened 2024-03-01
Update ploidy column and value

TCLamnidis:
Renamed ploidy column (one parameter for both freebayes and gatk).
Added ploidy of 1 to MT genome in reference sheet.


jfy133 (approved):
Mother of god that reference sheet...

PR #1110, opened 2024-03-01
Ribotish quality

pinin4fjords:
Add test data for riboseq. Specifically for ribotish/quality in this instance.


README.md
```diff
@@ -287,8 +287,12 @@ The folder is structured in the following way: Any nonspecific-pangenome file is
     - index
       - salmon: salmon index created with `transcriptome.fasta`
     - vep.tar.gz: Compressed VEP cache containing info.txt and synonyms of chr22 only. No annotations included.
+    - riboseq_expression
```
maxulysse, line 290: do we need a specific folder for riboseq, or can these files go where the other gtf and bam files are?

pinin4fjords: I wouldn't want people to use them like regular RNAseq or genomic alignments

maxulysse: sounds like a good enough reason for me

maxulysse (approved):

README.md
pinin4fjords: Thanks!

PR #1111, opened 2024-03-04
Add another alignment file for multi-input testing with ribotish/predict

pinin4fjords:
Just adding an extra test file to test the multi-bam inputs possible with ribotish/predict.


maxulysse:
Can you update the README?

PR #1121, opened 2024-03-14
Reduce funcscan testdatasets

Darcy220606:
This PR includes new test datasets for funsscan with a drastically reduced samplesize and renamed to match the samples id. The samplesheet was further updated.


jfy133 (changes requested):
We shouldn't change the paths in the existing samplesheet, thsi may break older versions of the pipelines/tests.

It would make more sense to have a new samplesheet (e.g. `samplesheet_reduced.csv`) that we can refer to in tests.

Please also briefly describe in a new section on the README how you ade these :)

jfy133 (approved):

PR #1122, opened 2024-03-18
taxprofiler: add a new test database including db_type

LilyAnderssonLee:
Add the new database test data, including the additional column `db_type`.


LilyAnderssonLee:
@jfy133 @sofstam In the new database test table, `kraken2` was duplicated with three different `db_type` values. There are no special `db_params` for the long database, so the `db_params` are the same.
- short  
- long  
- both

jfy133 (commented):
I think it would be good to have at least two 'long' databases (from two different tools), and some parameter within those to be slightly different - this would help with debugging.

Can be nonsense parameters, or simple things like changing kmer size or something

LilyAnderssonLee:
Actually, updating `kraken2`/`bracken`/`centrifuge` db is not a good option. Since there is only one Nanopore sample, it could lead to issues during the process of `STANDARDISATION_PROFILES` combining reports. Therefore, I duplicated the `kaiju` and `diamond` databases instead. I'm not sure if the `db_params` are broad enough. Please proceed with your suggestions.

LilyAnderssonLee:
It was a mistake, merging the master into the current branch. I reverted back now.

jfy133 (approved):
LGTM!

jfy133:
Although thinking about it, this might be a breaking change, so might have to bump to x2 (older input sheets will not work with this version of taxprofiler)

PR #1123, opened 2024-03-18
Add test for specific assembly id within taxa input

tillenglert:
The testfiles are used for a CI test with a fixed assembly id for a specific taxid


skrakau (approved):
It seems it allows for one input a mix of using fixed assemblies and just downloading the longest ones. Might be useful, a warning/note somewhere might be useful if not consistent. LGTM!

PR #1133, opened 2024-03-19
Test data for csv aggregation

josenimo:
Accounts for samples with or without sampleid column, if sampleid column not present the script will take the sampleid value from the stem of the filename (1.tif -> 1)

Created a new folder because the script expects a folders with all present CSVs to have a similar format, all other test dataset folders had markers.csv


kbestak:
Suggestion - exclude 2 of the input files. Subset data to be smaller.

Question, what is the Phenotyped and how is it different to Miguel's and Chiara's recently merged Phenotyped.csv?

PR #1140, opened 2024-03-19
Add test data for plink/ld

DavideBag:
Adding in popgen folder a random selection of snp to test linkage disequilibrium in module plink/ld.


lescai (approved):
LGTM

PR #1143, opened 2024-03-19
Add test-datasets for cobra-meta nf-core module

CarsonJM:


data/genomics/prokaryotes/bacteroides_fragilis/illumina/fasta/contig_id.tsv
sofstam, line 1: Should this be under a folder named `tsv` for example?

sofstam (commented):
Nice! One comment and I assume this [file](https://github.com/nf-core/modules/blob/master/tests/config/test_data.config) needs to be updated as well.

data/genomics/prokaryotes/bacteroides_fragilis/illumina/fasta/contig_id.tsv
CarsonJM: That seems like a good idea to me 👍

sofstam (approved):
Nice!

PR #1147, opened 2024-03-20
Plink simulated phenotype

LorenzoS96:
Plink file containing family IDs in the first column, within-family IDs in the second column and phenotype information in the third column


msantorsola (approved):
LGTM

PR #1153, opened 2024-03-20
Plink simulated phenotype file with header

LorenzoS96:
Plink simulated phenotype file with header to test the --pheno-name option. The file is a space-delimited file with the following columns:  Family ID ('FID')
Within-family ID ('IID')
Within-family ID of father ('IID_father')
Within-family ID of mother (''IID_mother')
Sex
Phenotype


lescai (approved):
this is just a simple added file so LGTM

PR #1167, opened 2024-03-26
Single-Cell Hashing demultiplexing

mari-ga:
Hi,
I'm currently working on the module DemuxEM for nf-core modules.
I have checked the existing datasets for the single-cell hashing demultiplexing tools and could not find any test datasets. Therefore, I would like to contribute with a small dataset, which comes from a single-nucleus RNA sequencing study, which is publicly available and was already used to test CellHashR/BFF (https://doi.org/10.1093/bioinformatics/btac213) one of the hashing demultiplexing modules.
It includes:

- One RNA 10x mtx matrix
- One HTO 10x mtx matrix
- One RNA H5 matrix
- One HTO CSV matrix


mahesh-panchal (approved):
Given this is not my field, the data looks smallish, and I'm not clear if this could be subsampled more without compromising anything. 

Please double check the guidelines here just in case I've missed something https://nf-co.re/docs/contributing/test_data_guidelines, but it looks like you've followed all these.

mari-ga:
Thank you!  yes, this dataset is already reduced to only two hashes for the 10x matrices, and the raw data is also small, especially in comparison with other available datasets.

PR #1169, opened 2024-03-28
Add dedicated 'metagenome' organism directory with metagenome related files

jfy133:
I'm increasingly coming across cases where the current test-data 'organism' structure does not work with metagenomic tools, such as the repeated `genome.fasta` files causing name conflicts.

Furthermore, as metagenomics are dealing with multiple organisms, I need combinations of multiple organisms e.g. with taxonomies.

This PR adds a dedicated metagenome folder that provides a space for 'minimal' metagenomes related files of just 2 organisms, including compatible names.

It also updates the README to include missing entries, where I know what they are...


README.md
```diff
@@ -560,6 +567,31 @@ The folder is structured in the following way: Any nonspecific-pangenome file is
     - nanopore
       - fastq
         - 'test.fastq.gz' synthetic raw long-read sequencing reads of the genome of the mammalian-gut-residing _Bacteroides fragilis_ YCH46 bacterium (NC_006347). Originally generated for the [MAG pipeline test dataset](https://github.com/nf-core/test-datasets/tree/mag).
+  - candidatus_portiera_aleyrodidarum
+    - genome
+      - ???
```
maxulysse, line 572: ???

README.md
```diff
@@ -560,6 +567,31 @@ The folder is structured in the following way: Any nonspecific-pangenome file is
...
+    - illumina
+      - ???
```
maxulysse, line 574: ???

README.md
```diff
@@ -560,6 +567,31 @@ The folder is structured in the following way: Any nonspecific-pangenome file is
...
+    - nanopore
+      - ???
```
maxulysse, line 576: ???

maxulysse (approved):

FriederikeHanssen (approved):
Looks very reasonable to me. We should just make sure somehow we don't have the fastas etc somewhere twice.

README.md
jfy133: I just don't know about them unfortunately... I vaguely remember they may have come from bactopia @rpetit3 ?

jfy133:
> Looks very reasonable to me. We should just make sure somehow we don't have the fastas etc somewhere twice.

These are technically there twice :\, but `gneome.fasta` twice  doesn't work well when trying to stage them together in one directory :grimacing:

jfy133:
Ok I have the two :heavy_check_mark: so proceeding even if it's sarek focused :eyes:

PR #1173, opened 2024-04-05
Add barcodes.tsv file required to run test dataset with scsplit

IreneRobles:
Hi, 

I am developing the `scsplit` module. I modify my tests to use the same dataset as `popscle`, but I need an extra `barcodes.tsv` file. I tried some `barcode.tsv.gz` test datasets present in the branch, but they do not run, so I decided to add one that does.

Have a nice day,

Irene


data/genomics/homo_sapiens/demultiplexing/README.md
wxicu, line 1: Cell barcodes are also provided by popscle (Source: https://drive.google.com/file/d/1n52EZFz83W6lNN8jo8r1hQqVoTipLhoT/view?usp=drive_link)

wxicu (changes requested):
Hi @IreneRobles  Thanks for adding the file. However the barcode file is also provided by popscle, not provided by hadge.

data/genomics/homo_sapiens/demultiplexing/README.md
IreneRobles: Now the cell barcodes are from popscle, :)

IreneRobles:
Hi @wxicu ,

I switched the barcodes. Now they are from popscle. :)

jfy133 (commented):
You also need to update the main README of the whole repo (well on that branch), describing how the file was generated not just the demultiplex specific README.

Also the file is very large at 12MB, is there anyway it could be made smaller?

IreneRobles:
Hi @jfy133 and @wxicu ,
- I have reduced the  size of `barcodes.tsv` file so it only contains the barcodes of the cells in `chr21.bam`
- I have also updated the `README.md`s
Thank you for taking a look at my branch!

jfy133 (approved):

jfy133:
One last thing now I thin of it: It looks like the barcodes might be specific to a particular library build.technology? If yes maybe `barcodes.csv` could be renamed to reflect that.

If they are generic barcodes it's probably fine :)

wxicu (approved):

PR #1178, opened 2024-04-12
Add barcode and bam index file for demultiplexing

wxicu:
Add barcode and bam index file for demultiplexing modules (https://github.com/nf-core/modules/issues/5072)


wxicu:
duplicate with #1173 closed

PR #1185, opened 2024-04-17
Change pipeline name to nf-core/spatialvi

fasterius:
As per discussions in the nf-core community, this commit changes the name of the pipeline from `nf-core/spatialtranscriptomics` to `nf-core/spatialvi`. Relevant PR is at https://github.com/nf-core/spatialtranscriptomics/pull/77.

This PR also updates the documentation to reflect the changes previously introduced, but where the documentation was not updated.

Not sure if (1) merging this PR into the existing `spatialtranscriptomics` branch and renaming it, or (2) creating a new branch with the `spatialvi` name and deleting the old one is the best way to go. Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ﻿# ![nfcore/test-datasets](docs/images/test-datasets_logo.png)
 
 This branch contains test datasets as well as various files for CI and unit
-testing for the [`nf-core/spatialtranscriptomics`](https://github.com/nf-core/spatialtranscriptomics)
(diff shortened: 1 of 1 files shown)


maxulysse:
How about creating a new `spatialvi` branch out of `spatialtranscriptomics`, merging said PR in it, and keeping the old one for reproducibility purpose if people are still using older versions of the pipeline.

fasterius:
Excellent idea! I created a new branch and changed the target. I'll also do a separate PR for the documentation updates to the original branch, so that it's accurate.

PR #1187, opened 2024-04-22
added transcriptome bams for riboseq

iraiosub:
added transcriptome bams for riboseq - required for riboWaltz module


pinin4fjords (changes requested):
Could you update the README? You'll see the existing bams listed there.

Also, the naming of these suggests to me that they're the sorted files from the riboseq test profile run. I've actually found issues using those with e.g. Salmon, the sorting makes less sense for transcriptome BAMs (it's just what the relevant subworkflow does right now). 

Maybe it would be simpler to use the unsorted ones unless Ribowaltz needs the sorted ones?

iraiosub:
thanks for your feedback @pinin4fjords ! I've made the requested changes. Could you please take another look and see if everything is now in order?

pinin4fjords (changes requested):
Thanks for the changes! Just need to update the top-level README as well for the new files, and you're done.

iraiosub:
done! next time will know what to change from the first go:). thanks!

pinin4fjords (approved):
Good work!

PR #1193, opened 2024-04-26
added tar.gz of database for eager2 test profile

ilight1542:
This PR is to add a tar.gz version of the test database so that it is retrievable for dsl2-eager malt testing profile


jfy133 (commented):
Need docs in the README of what it is/how it was generated

jfy133 (approved):

PR #1196, opened 2024-05-06
Add KrakenUniq databases

jfy133:
And apparently includes some prettier formatting...?

Built with the following steps

1. Downloaded from `sarscov2` and `prokryote/metagenome ready:
    1. FASTA files you want to include
    2. A taxonomy `.map` file that contains ``<FASTA_ACCESSION_ID>\t`TAX_ID>
    3. `nodes.dmp` taxonomy file
    4. `names.dmp` taxonomy file
2. Make a directory named with the name the database will be called
3. Within this directory, make two further directories
    - `library/`
    - `taxonomy/`
6. Within these two directories place the following
    - `library/`: All fastas and seq map file
    - `taxonomy/`: the two `*.dmp` files
9. Run `krakenuniq-build --db <db_name>`
    - You may need to include `--jellyfish-bin $(type -P -a jellyfish)` if you get an unbound variable warning
10. If successful, should have a bunch of files in `<db_name>` like `.kbd`, `.jdb`, `.tsv` `library-fioles.txt`, `taxDB` etc.
11. Run a test: ` krakenuniq --db customdb_krakenuniq_mini/ --threads 4 --quick test_1.fastq.gz`
12. If all good Run cleanup: `rm -r *log *jdb *counts *tsv *txt *map libraries/ taxonomy`

Tested by running both FASTA and FASTQ files:

```
$ krakenuniq --db krakenuniq/ --threads 4 --quick contigs.fasta
/home/james/bin/miniconda3/envs/krakenuniq/share/krakenuniq-1.0.4-1/libexec/classify -d krakenuniq//database.kdb -i krakenuniq//database.idx -t 4 -q -a krakenuniq//taxDB -p 12
 Database krakenuniq//database.kdb
Loaded database with 29799 keys with k of 31 [val_len 4, key_len 8].
Reading taxonomy index from krakenuniq//taxDB. Done.
C	NODE_1_length_20973_cov_191.628754	2697049	20973	Q:1
C	NODE_2_length_8774_cov_178.827802	2697049	8774	Q:1
U	NODE_3_length_5473_cov_3.280771	0	5473	Q:0
U	NODE_4_length_458_cov_1.051360	0	458	Q:0
C	NODE_5_length_446_cov_1.078370	2697049	446	Q:1
5 sequences (0.04 Mbp) processed in 0.016s (18.7 Kseq/m, 135.41 Mbp/m).
  3 sequences classified (60.00%)
  2 sequences unclassified (40.00%)
```

```
[...]

PR #1197, opened 2024-05-06
Reduce size of KU database

jfy133:
Same as #1196 but with smaller kmer and minimzer lengths ((`--kmer-len 15 --minimizer-len 13`)


Midnighter (approved):
🎉

PR #1198, opened 2024-05-08
Fix sample group to match sample filename

Aratz:
During testing it was noticed that the group of one sample didn't match the group in the filename of one sample.

This PR fixes this issue.


maxulysse (approved):

Aratz:
Thanks @maxulysse !

PR #1200, opened 2024-05-09
initial testdata for nf-core/fastquorum

nh13:


jfy133 (approved):
Based on the specifications here: https://nf-co.re/docs/contributing/nfcore_component-specifications/test_data

This overall LGTM. Ideally it would be nice if the FASTQ files were a little smaller (if possible), but otherwise it's within bounds.

Only suggestion is given you've given the commands used to generate data for completeness the versions  of `seqtk`, `samtools`, and `bwa`, would be nice (thinking something got nuked and had to be re-generated), but not essential.

maxulysse (approved):
<3

PR #1205, opened 2024-05-17
Third FASTQ containing UMI sequences

adamrtalbot:
This PR adds a third FASTQ file which contains UMI sequences, which matches the read structure of some library preparation kits which embed the UMI in an index sequence.

It adds the third UMI to the samplesheet for testing when running fastquorum in `-profile test` mode. Question for @nh13, shall we create a distinct samplesheet for running this use case? When combined with nf-test we can test a range of inputs.


famosab:
@adamrtalbot Bump :) Do you still need this?

nh13:
Closing this as this is likely a bit stale, but happy to take re-open if folks are interested!

PR #1219, opened 2024-05-22
Adding the test datasets for ViennaRNA/RNAcofold and ViennaRNA/RNAup

kokul-atx:
This test data set will be used to test the ViennaRNA tools for nf-core/modules that require hybridization.


FloWuenne (approved):
LGTM ⭐ !

PR #1224, opened 2024-06-04
modkit pileup test data

fellen31:
Putting a phased copy of the reads in `data/genomics/homo_sapiens/nanopore/bam` in `delete_me`, since I'm not sure they are human (they are aligned to CGI_ladder_3.6kb_ref.fa).

Can be removed when there's (ideally):
- [ ] A module that can phase long-reads and
- [ ] ONT sarscov2 data with methylation


jfy133:
@fellen31 can you identify from the commit that added the original bam to ask what reference it is? Ideally you should have phased human reads in the human directory so others can use it and not hide it in delete me...

fellen31:
@jfy133 That's fair. I've put more general (aligned & phased) ONT testdata aligned to `genome.fasta` instead, more inline with the other/illumina test data. This should be more useful for others as well.

- test.sorted.bam: 24 reads from HG002_R1041_UL_dorado0.4.0_sup4.1.0_5mCG_5hmCG sorted and mapped to genome.fasta (chr22:16570000-16610000)
- test.sorted.bam.bai: Index for test.sorted.bam
- test.sorted.phased.bam: Haplotagged version of test.sorted.bam
- test.sorted.phased.bam.bai: Index for test.sorted.phased.bam

SPPearce (approved):
That seems much more useful, great work

PR #1238, opened 2024-06-17
Update values of db_type in the test databases

LilyAnderssonLee:
Replace the `db_type` value `both` with `short,long`.


jfy133 (approved):
Hrm, maybe the comma separated notation is a bad idea because we need the quotes but this works for now 👍

LilyAnderssonLee:
Good point. Maybe use `; `instead? I can update both here and the split_db [PR](https://github.com/nf-core/taxprofiler/pull/466).

jfy133:
Yes! good idea!

PR #1239, opened 2024-06-24
add ascat loci and allele files

famosab:


maxulysse:
Do we really need all that for tests?

famosab:
Maybe one chromosome is enough? Pinging @mauro-saporita: Mybe you can try to construct a test case only with chr21

mauro-saporita:
It looks like we really need the entire folder with all chr.

famosab:
@mauro-saporita whats the status here?

mauro-saporita:
> @mauro-saporita whats the status here?

Hi @famosab, sorry for late response, I was on leave in the last couple of months. 
No updates on this PR - I will do my best to look into it in the next weeks.

famosab:
I'll close this see https://github.com/nf-core/modules/pull/7727

Issue #1240, opened 2024-06-25
Update test data for bcl_demultiplex subworkflow


k1sauce:
Illumina's BCL Convert software generates FASTQ files for sequencing data run directory containing BCL files. The demultiplexing process involves assigning reads to a FASTQ files as specified by the indices (barcodes) in the SampleSheet.csv file. If no reads can be assigned to the barcode, bcl-convert generates an empty FASTQ.  The empty FASTQ file has caused error in the bcl_demultiplex subworkflow. The fix should include a test case for empty FASTQs. We can include a sample with indices in the SamleSheet.csv that are not used in the library prep to generate the empty FASTQ file during the test.

PR #1241, opened 2024-06-25
feat: add indices to generate an empty FASTQ with bcl-convert

k1sauce:
Adds indices to generate an empty FASTQ when running bcc-convert. This addresses #1240


Aratz (approved):
Thanks for addressing this issue!

PR #1276, opened 2024-08-06
Staging phenoimager

chiarasch:
Add raw tiles from PhenoImager for testing staging module staging/phenoimager2mc. Can I somewhere add the reference https://data.mendeley.com/datasets/5vfz9vhm2s/1 of the data subset?


kbestak (approved):
LGTM, especially since these are the two smallest possible neighoring tiles in the dataset

PR #1287, opened 2024-08-15
Create samplesheet_technical_repeats.csv

atrigila:
Add a samplesheet with "technical repeats" (same sample name) to add testing for issue https://github.com/nf-core/smrnaseq/issues/212


apeltzer (approved):
Makes sense :)

PR #1292, opened 2024-08-21
Cafe

chriswyatt1:
I have made a super tiny dataset (8k), as there are no equivalents on test-datasets already. 

I have briefly described the contents


SPPearce:
Mmm, this should have been into the modules branch, rather than into a new branch on its own.

PR #1295, opened 2024-08-23
Add tags to input samplesheet

Aratz:
To allow for a more flexible way to group samples into various report, we'll adopt a tag system, rather than using specific columns (such as `lane` and `group`).

Relates to https://github.com/nf-core/seqinspector/issues/16


maxulysse (approved):

maxulysse:
I like this tag system

PR #1297, opened 2024-08-24
Revert "Cafe"

chriswyatt1:
Reverts nf-core/test-datasets#1292

This should not have been saved here, but in the modules branch


FernandoDuarteF (approved):
All good

PR #1300, opened 2024-08-26
add bowtie2

Joon-Klaps:
Adding bowtie2 index for testing.


matthdsm (commented):
Shouldn't this be generated on the fly using nf-test `SETUP`?

maxulysse (dismissed):

maxulysse:
Actually good comment by @matthdsm

Joon-Klaps:
> Shouldn't this be generated on the fly using nf-test `SETUP`?

It should, when its using nf-test. Unfortunately, viralrecon isn't there yet. Additionally (and the reason why I made this PR), is that viralrecon has a param `bowtie2_index` that was crashing the pipeline.

Joon-Klaps:
Could of course also remove the (kinda redundant) param...

matthdsm:
Not sure what the best way to go here is 🤷🏻‍♂️

PR #1305, opened 2024-08-28
smraseq - add nextflex test data

lpantano:
add test data for next flex


lpantano:
@ apeltzer, is this ok for a test data? it is coming from subsampling of this dataset https://github.com/nf-core/smrnaseq/issues/366

PR #1310, opened 2024-08-31
Orthofinder test data (zipped)

chriswyatt1:
Added a tar.gz folder as test data for orthofinder, instead of a unzipped one. 
Now the orthofinder test will work, as it untars this in the setup of the test


FernandoDuarteF (approved):
All good

PR #1312, opened 2024-09-02
Add test data to develop nf-core `mirtop` module

atrigila:
Add data to test nf-core `mirtop` module
Related to: https://github.com/nf-core/smrnaseq/issues/403


nschcolnicov:
@atrigila  Looks good, just one question, I see that you placed it in a folder called "delete_me", so I imagine these are just temporary files. Why are they temporary?

PR #863, opened 2023-04-24
Salmon results compressed and the updated samplesheet for "test_salmon" configuration

valentinoruggieri:
[ERR188383.tar.gz](https://github.com/nf-core/test-datasets/files/11310269/ERR188383.tar.gz)
[ERR188428.tar.gz](https://github.com/nf-core/test-datasets/files/11310270/ERR188428.tar.gz)
[ERR188454.tar.gz](https://github.com/nf-core/test-datasets/files/11310271/ERR188454.tar.gz)
[ERR204916.tar.gz](https://github.com/nf-core/test-datasets/files/11310272/ERR204916.tar.gz)
[samplesheet_salmon.csv](https://github.com/nf-core/test-datasets/files/11310281/samplesheet_salmon.csv)

PR #707, opened 2022-12-01
Add rnasplice test data

bensouthgate:

PR #1314, opened 2024-09-04
rnasplice test data

lathikaa:
Previously there was a branch 'rnasplice' in nf-core test datasets. https://github.com/nf-core/test-datasets/pull/707

This was deleted recently(https://github.com/nf-core/test-datasets/pull/863), so raising a pull request again as the nf-core/rnasplice pipeline will not work without this test data.


jma1991:
@lathikaa I have merged a PR for rnasplice separately (#1332)

PR #1317, opened 2024-09-05
Add Nanopore fasta entry to the samplesheet

LilyAnderssonLee:
Related to https://github.com/nf-core/taxprofiler/pull/518


jfy133 (approved):
I would maybe make it 2110 and go on the first line to group similar samples, but yes this is what we want :)

PR #1321, opened 2024-09-11
small hifi dataset for flye

famosab:


fellen31 (changes requested):
Could you add this to the README? 

(perhaps name it `small_data` instead of `small-data`? Might just be personal preference, but I think most other files use _ 😄)

famosab:
Its from here: https://www.pacb.com/connect/datasets/#AAV-datasets
I think I need to rename and properly document what is in this file. I'll re-request the review then :)

fellen31:
> Its from here: https://www.pacb.com/connect/datasets/#AAV-datasets I think I need to rename and properly document what is in this file. I'll re-request the review then :)

Just a link to the download file and a very short description in the modules README is okay for me.

famosab:
This is not needed anymore since we decided to skip the `--raw` tests

PR #1334, opened 2024-09-23
evexplorer

ammarsabir15:
I have added a test-dataset for evexplorer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,5 +3,9 @@
 
 # Content of this repository
 ## ev_small_rna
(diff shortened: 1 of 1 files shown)


jfy133 (commented):
I've switched your PRto point to a dedicated `evexplorer` branch, not `master` as each pipeline has their own.

Otherwise: is this public data? It's important that we have stuff we are 'allowed' to use (e.g. not contain any host/sensititve/private data); if you do have permission please put a statement who gave permission/where from), otherwise if it is public data please list where you got it from :)

Also if it's a subsampled file - it's good to describe how

jfy133 (approved):

PR #1335, opened 2024-09-24
Update cell ranger - hashing demultiplexing data

mari-ga:
* Update data to include the third column generated by Cellranger version 3 for compatibility with packages such as Scanpy, in order to solve this error when opening the data:
https://github.com/scverse/scanpy/issues/1916

* New data includes a third column in the `features.tsv.gz` files in both hto and rna directories.


nvnieuwk:
How will this affect tests that are currently using these files?

PR #1338, opened 2024-10-02
pypgx testing data 

Jorisvansteenbrugge:
Hi,

I added a small temporary test dataset in data/delete_me on the modules branch to test multiple new nf-core modules (e.g., https://github.com/nf-core/modules/pull/6720) for the tool PyPGx. PyPGx requires reads to be mapped to a different region than is currently represented in the human illumina bam files in the test-dataset repository. I have started a thread on Slack to discuss the best way to test this tool. However, to avoid delaying the creation of the modules, I have uploaded some temporary test data to be used in the meantime.

### DIFF
--- a/data/genomics/homo_sapiens/README.md
+++ b/data/genomics/homo_sapiens/README.md
@@ -391,7 +391,7 @@ This dataset contains:
       -  22:42512500-42551883
       -  22:42512500-42551883
    - illumina/bam/test.PGx.CYP2D6.bam.bai => The index of the BAM file
-   - genome/GRCh37_chr22.fasta.gz => A reference GRCh37 genome fasta file (bgzipped) containing chromosome 22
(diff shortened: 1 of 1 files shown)


jfy133 (commented):
I would suggest moving these into the homo_sapiens folder, as delete me is not sable.

Please also update the branch's README descriving how the files were generated etc.

Jorisvansteenbrugge:
@jfy133 Thank you, I moved the data to the homo_sapiens folder, and I updated the README in that same folder. Just to be sure, is that the README you were referring to?

jfy133:
Yes, but also in the main module README.

I don't think you need a specific `pypgx` directory though, the files can go in `genome` and `illumina` folders respecively, the formats are not specific to pypgx

Jorisvansteenbrugge:
@jfy133 oke perfect, your comments should be addressed now!

jfy133 (approved):
Normally I like to also say how it was mapped (tool, version etc) but I think I've asked enough now 😬

There is something in all the places that it needs to be now

PR #1344, opened 2024-10-11
Add bcf.gz copy

fellen31:
The only compressed BCF available is `data/genomics/homo_sapiens/popgen/plink_simulated.bcf.gz`, I need a file with only one sample for a new deepvariant/vcfstatsreport module.


Lucpen (approved):
LGTM!

PR #1350, opened 2024-10-17
createtaxdb: add broken samplesheets

jfy133:


nvnieuwk (approved):
LGTM! Where will this be tested?

jfy133:
Manually for now, but if I work out a nice way (with nf-test maybe) can add that in the future

nvnieuwk:
Allright, I'll probably add them to the nf-schema tests themselves then if you are okay with that :p

jfy133:
> Allright, I'll probably add them to the nf-schema tests themselves then if you are okay with that :p

Go ahead!

PR #1358, opened 2024-10-24
add test files for genomassembler

nschan:
Hello,

this adds test files for the genomeassembler pipeline.

The files are ONT (10.14) and pacbio HiFi reads that assemble to A.thaliana Col-0 Chr1 region 5MB-7MB.


mahesh-panchal (commented):
Can you put them in a folder with a README describing a little bit about the test data set like how it was generated and the source of the data please.

nschan:
Something like that, or do would you like additional information? The data has not been published.

mahesh-panchal (approved):
It would be nice to have an accession, but you make a PR later. I don't want to hold up inclusion because of that. 

Can you change the ending of the README file to .md please?

PR #1361, opened 2024-10-25
Typo in input_myco_tiny.csv

FernandoDuarteF:
There was a typo in ``input_myco_tiny.csv`` which didn't allow file access.


mashehu (approved):

chriswyatt1 (commented):
Yes, but with the new fastq column

PR #1363, opened 2024-10-28
Modules

tanyasarkjain:
Test data for PBSV/call (Pacbio structural variant caller) which requires an intermediate svsig file - signatures of structural variants - required for this module https://github.com/nf-core/modules/pull/6818 (the call submodule)


fellen31 (approved):
Another option is to use `setup` in nf-test, to be able to run the pbsv_discover module before testing pbsv_call. For a simple example: https://github.com/nf-core/modules/blob/06fa1219e57fe1b2154e5ec966ddbff3e1c7daba/modules/nf-core/bwa/aln/tests/main.nf.test#L14

PR #1366, opened 2024-10-29
update samplesheets for genomeqc

stephenturner:
* bca16674 create new samplesheet using refseq accessions but with an empty fastq column for testing, see genomeqc#42 genomeqc#62 
* 2029c019 - add emtpy fastq column for fasta+gff samplesheet see genomeqc#42 genomeqc#62 (2 minutes ago)

See also, previous (accidental) direct push to this branch, 5e7f535, which adds fastq files directly to the refseq samplesheet


chriswyatt1 (approved):
Looks good to me, this can be merged

PR #1397, opened 2024-11-21
update mgikit dataset

ziadbkh:


JoseEspinosa (changes requested):
LGTM, but would be nice if you could update the readme as I mentioned on your previous PR [here ](https://github.com/nf-core/test-datasets/pull/1396#pullrequestreview-2450839529) (after merging though 😛)

JoseEspinosa (approved):
Thanks a lot @ziadbkh! 🚀

PR #1407, opened 2024-11-27
Add Panel Of Normal and NA12878 data

LouisLeNezet:
I've added here the different files necessary to perform phasing and imputation using the 1000 Genome Project and the data from the individuals NA12878.
The limitation are only the chromosome 22 is used and a second individuals would be nice to test batch imputation (example in GLIMPSE2_PHASE.


kubranarci (approved):
Seems good to me 👯‍♀️

PR #1411, opened 2024-12-03
Fix name of NA12878

LouisLeNezet:
Here I changed the name of the individual to NA12878 (was HG001 before).
I also compress the map file and add a 1X imputed file for this individual for the chr22.


data/genomics/homo_sapiens/genome/genome.GRCh38.chr22.map
maxulysse, line 1: was that file used before in any test?

LouisLeNezet: Not yet, I've added it yesterday

maxulysse (approved):

PR #1416, opened 2024-12-06
fix chr name in reference fasta

alanmmobbs93:
Remove the coordinates from the sequence name in order to match the gtf file


famosab:
@alanmmobbs93 Do you still need this update? :) Feel free to mark this as ready for review then or if not needed close the PR!

PR #1417, opened 2024-12-06
Update fasta

nschcolnicov:


nschcolnicov:
Difference with previous version is just: 

image

Then the index was re-generated

PR #1423, opened 2024-12-11
Add fusionreport data to rnafusion test-datasets

atrigila:
https://github.com/nf-core/rnafusion/issues/556


alanmmobbs93 (approved):
Like it! Just one question, is the HTML file necessary?

atrigila:
yes

PR #1437, opened 2024-12-17
Add `CNVKIT` test data

LouisLeNezet:
Here is added the test data for the `CNVKIT` processes.
They are from the [cnkit](https://github.com/etal/cnvkit/blob/master/test/formats/) repository.
But will need to be interoperable with the rest of the nf-core test datasets before being put in the main folder.
Therefore they are put in the `delete_me` folder.


LouisLeNezet:
Okay, make sense also for me.
I will try to use the CNVKIT modules to generate correct files.

SPPearce (approved):

PR #1446, opened 2024-12-22
Removed too small files and added test2_1 corrupted file for gzrt and …

mazzalab:
- Removed useless test data because too small, i.e., a very small number of recoverable reads inside
- Added a 10 kb file that is the corrupted version of test2_1


sateeshperi:
update the README accordingly

mazzalab:
@sateeshperi done, sorry

sateeshperi (approved):

PR #1454, opened 2025-01-09
Add SKA sketch files for SKA/DISTANCE

itrujnara:
Added SKA sketch files created from SARS-CoV-2 and P. falciparum genome fragments. They are valid SKA input but have no biological meaning.


JoseEspinosa (approved):
Thanks! LGTM 🚀

PR #1465, opened 2025-01-16
Added missing file and updated readme

nschcolnicov:


SPPearce (approved):

data/genomics/mus_musculus/rnaseq_expression/SRP254919.contrasts.yaml
```diff
@@ -0,0 +1,6 @@
+contrasts:
+  - id: treatment_mCherry_hND6_
```
SPPearce, line 2: Is this missing something at the end?

pinin4fjords: No, it's an empty string

PR #1466, opened 2025-01-21
Add transcriptome.fa syncd with the GTF

pinin4fjords:
The current transcriptome.fa has a Gfp sequence in it, which does not match the GTF. This is because, in current testing, the GTF is augmented during the cat_additional_fasta process by which a reference fasta sequence is augmented by an 'additional' one. 

However, I'm now adding a test for a genomic fasta-less mode of operation, in which that step will not happen, and the GTF will therefore remain out of synch. I need a transcriptome fasta matching the GTF, without that Gfp.


FriederikeHanssen (approved):

pinin4fjords:
Thanks @FriederikeHanssen !

PR #1483, opened 2025-01-31
Extra dataset for genomeqc

chriswyatt1:
Added a new dataset to the raw data, as we need three species to run the orthofinder module


chriswyatt1:
It works testing it. 
We still need the conf file in the genomeqc repo, and test once the files are pulled from there

FernandoDuarteF:
Cool! I'll add the config file.

PR #1485, opened 2025-01-31
Msa toolsheet v1.1

luisas:
Remove LEARNMSA from standartd testing but adding it for the release full test ( with the right flags!)


mirpedrol (approved):
LGTM 🚀

PR #1495, opened 2025-02-10
Update titanic_model.py

itrujnara:
This should fix nf-test issues. Some lines were missing for some reason.


mathysgrapotte (approved):
it looks good for me

PR #1500, opened 2025-02-17
add embl sarscov fasta

famosab:


famosab:
Ah ok then the description is just wrong but the file that I submitted is the one I need haha. I can adjust that in the readme?

vagkaratzas:
> Ah ok then the description is just wrong but the file that I submitted is the one I need haha. I can adjust that in the readme?

If it's for a specific module that can't run with the existing fasta files then sure. Just rename to something like this:
`'genome-ena.fasta': ENA associated fasta sequence`

If it's pipeline specific, then just move it to the pipeline's test-datasets branch

PR #1507, opened 2025-02-28
Add Metaphlan test-dataset (WIP nf-core/eager#1085)

merszym:
Add a BAM file + corresponding samplesheet that produces an output with MetaPhlan4 and the test-database in this repository


ilight1542:
LGTM

ilight1542 (approved):
Later we should create also a test dataset for each tool when skippairmerging is on (I can take the lead on that next week)

PR #1513, opened 2025-03-05
 add somalier ancestry references from 1000 Genomes project

rannick:
add somalier ancestry references from 1000 Genomes project


fellen31 (commented):
If this is about the smallest test-data set that the module works with, I think it's better to not have it in `delete_me`.

A line in the README would also be nice!

SPPearce:
Yes, please don't put more stuff into `delete_me`!
Can the file be any smaller? It is still 15Mb.

famosab:
@rannick Small bump :)

rannick:
We realized that somalier ancestry is performing worse than peddy (https://github.com/brentp/somalier/issues/103) so I will close this

PR #1519, opened 2025-03-10
Add dream dataset

nschcolnicov:
Adding a required dataset for this modules PR: https://github.com/nf-core/modules/pull/7750
@apeltzer @grst


grst (approved):
The dataset is fine, but why do we even need a new one for DREAM specifically?

nschcolnicov:
> The dataset is fine, but why do we even need a new one for DREAM specifically?

Its just that I tested the pipeline with this dataset, and I already know this one work

PR #1523, opened 2025-03-14
Add bam csi indexes to test graphtyper/genotype

zachary-foster:
These are needed by `graphtyper/genotype`. I want them to make tests pass so I can submit an update to `graphtyper/genotype`.

Needed for https://github.com/nf-core/modules/pull/7804


famosab:
Can you add the data to the README please? :)

zachary-foster:
Ok, I added them. Also, it looked like that section of the readme was out of date, so I updated the other files there too to match what is in the directory.

PR #1536, opened 2025-03-24
Add MiSeq data for full tests

Aratz:
This PR adds two MiSeq datasets: one with single end reads, and one with paired end reads.

The data was downloaded from demo datasets on basespace (run ids `288644377` and `291339102`) and demultiplexed with nf-core/demultiplex

Since the datasets are not that big (both < 100MB), I was thinking they could be hosted in this repo? Feel free to disagree and help me to upload them to the s3 bucket instead.


maxulysse (approved):
LGTM

PR #1540, opened 2025-03-25
Corrected Samplesheet and Additional Samples

KarNair:
- Added One more sample
- Corrected Samplsheet to be in line with the latest format
- Updated Readme


Aratz (approved):
Thanks! :pray:

PR #1548, opened 2025-03-25
Add VueGen module test data

sayalaruano:
I created a folder with expected files for the VueGen module and added it to the delete_me folder for now, to be changed with available data in the future.


famosab:
1. Can you add this to the README please?
2. Why do we need both the tar and the extracted version? I would just use the tar and then always have gunzip as setup when testing (saves us a lot of space!)

jfy133:
I agree with the tarred version, you can't 'pull' whole directories via HTTPS in tests

sayalaruano:
Thanks for the suggestion @jfy133 and @famosab. Hopefully, it's ok now :)

jfy133 (approved):

famosab:
I am sorry but you also need to add this to the overview README (at least thats my opinion). Or at least that is what I meant.

jfy133:
> I am sorry but you also need to add this to the overview README (at least thats my opinion). Or at least that is what I meant.

Delete_me is the exception here, so it's not necessary - unless people have started doing it?

PR #1550, opened 2025-03-25
Fix samplesheets for full tests

Aratz:
The link to the undetermined fastq file was wrong in the single end samplesheet.

We also decided we could combine both samplesheets into a single test.


full_tests/MiSeq_SingleEnd/SampleSheet.csv
```diff
@@ -1,5 +1,5 @@
 sample,fastq_1,fastq_2,rundir,tags
-Undetermined,miseq_single/Fastq/Undetermined_S0_L001_R1_001.fastq.gz,,miseq_single,
+Undetermined,https://github.com/nf-core/test-datasets/raw/seqinspector/full_tests/MiSeq_SingleEnd/Fastq/Undetermined_S0_L001_R1_001.fastq.gz,,,https://github.com/nf-core/test-datasets/raw/seqinspector/full_tests/MiSeq_SingleEnd/rundir.tar.gz,
```
KarNair, line 2: Seem like this line has an extra comma either. Either in the tag or in after fastq_1.

KarNair (changes requested):

KarNair (approved):
LGTM!

PR #1551, opened 2025-03-25
Update databasesheet for v.1.3

sofstam:


database_full_v1.3.csv
```diff
@@ -0,0 +1,19 @@
...
+kraken2,kraken2-db,,short,s3://ngi-igenomes/test-data/taxprofiler/kraken2.tar.gz
+krakenuniq,krakenuniq-db,,short,s3://ngi-igenomes/test-data/taxprofiler/krakenuniq.tar.gz
+malt,malt-db,,short,s3://ngi-igenomes/test-data/taxprofiler/malt.tar.gz
+metaphlan,metaphlan3-db,--mpa3,short,s3://ngi-igenomes/test-data/taxprofiler/metaphlan3.tar.gz
+motus,motus-db1,,short,s3://ngi-igenomes/test-data/taxprofiler/db_mOTU.tar.gz
+motus,motus-db2,,long,s3://ngi-igenomes/test-data/taxprofiler/db_mOTU.tar.gz
+motus,motus-db3,,short;long,s3://ngi-igenomes/test-data/taxprofiler/db_mOTU.tar.gz
+ganon,ganon-db,,short,s3://ngi-igenomes/test-data/taxprofiler/ganon.tar.gz
+kmcp,kmcp_db,,short,s3://ngi-igenomes/test-data/taxprofiler/kmcp.tar.gz
+sylph,sylph_db,,short;long,https://github.com/nf-core/test-datasets/raw/taxprofiler/data/database/sylph/database.syldb
```
LilyAnderssonLee, line 19: Why is the database type `short;long` instead of `long`?

sofstam: Because sylph can work with both short and long reads. But maybe this column is not for that?

LilyAnderssonLee, line 19: Should the `sylph` database be compressed, with extension `tar.gz`?

sofstam: I have been following the same format as we have for diamond.

LilyAnderssonLee: Ok. If the database file isn't so big, then it's fine without compression.

LilyAnderssonLee (approved):

PR #1570, opened 2025-03-31
Update sylph taxonomy

sofstam:


jfy133 (approved):

jfy133:
Any docs update required?

sofstam:
Yes! So, far it has been bash scripting. I think it would be smoother if we can get this taxonomy with `https://github.com/shenwei356/gtdb-taxdump` and update the docs accordingly.

PR #1581, opened 2025-04-08
Add new inputs to rnafusion

nvnieuwk:
These will be used to test BAM/CRAM/junctions/splice_junctions input files in the pipeline


atrigila (approved):
Could you add a README for reference on how you obtained these files / how these can be reproduced? Thanks!

PR #1582, opened 2025-04-08
Modules

delfiterradas:
Add test data for starfusion module: https://github.com/nf-core/modules/pull/8107


famosab (changes requested):
Can you add your data to the README aswell? :)

famosab (approved):

PR #1591, opened 2025-04-15
Update and reorganize pixelator test data

johandahlberg:
This updates and reorganizes the test data for the `nf-core/pixelator` pipeline. I am splitting the data into two main folders, one for Molecular Pixelation data (`mpx`) and one for the Proximity Network Assay (`pna`). This can then be used to test the different paths through the nf-core/pixelator pipeline.


fbdtemme (commented):
I think we should keep the toplevel structure the same just to keep it similar to the other pipelines.

- ./samplesheet
- ./testdata

And duplicate the pna/mpx split in those two dirs.

That will also make it a bit easier to keep the current tests running since we wont have to sync up the changes to the testdata there.

fbdtemme (approved):

PR #1601, opened 2025-04-29
Add sequencing platforms to test samplesheets

muabnezor:


jfy133 (changes requested):
We should not add new columns to existing sheets in case it breaks testing of older version

Can you make duplicates each of the samplesheets and suffix them with `v4` or something (your long reads PR will be released with mag v4).

You can also take the opporunity to move the columns to after group and before short_reads_1 (which is where intuitively I would expect the column to be)

muabnezor:
> We should not add new columns to existing sheets in case it breaks testing of older version
> 
> Can you make duplicates each of the samplesheets and suffix them with `v4` or something (your long reads PR will be released with mag v4).
> 
> You can also take the opporunity to move the columns to after group and before short_reads_1 (which is where intuitively I would expect the column to be)

All right! We can also wait with these changes until v4 is in place.

muabnezor:
added .v4 suffix to test samplesheets for mag v4, and kept old samplesheets

jfy133 (approved):
Please add a sentence describing these collection of files (the main difference with the previous one) on the README, then LGTM!

PR #1603, opened 2025-05-05
CREATETAXDB: Add custom malt `.db` file

jfy133:
So tests run faster (otherwise have to download a ~1GB file from MALT/MEGAN website)


atrigila (approved):
Looks good!

PR #1625, opened 2025-05-28
feat: add gatk config for glnexus as a yml file

awgymer:
Adding a YML file with the default gate config for glnexus so we can test custom config use in the nf-core module


charles-plessy (approved):
Assuming you have local tests working well with the copy in your [awgymer:add-glnexus-config](https://github.com/awgymer/nf-core-test-datasets/tree/add-glnexus-config) branch, the PR looks good to me (and to ChatGPT…)

PR #1626, opened 2025-05-28
Add universc reference genome

LouisLeNezet:
Add the universc reference genome processed


SPPearce (changes requested):
Can you add it to the top level README as well please.

LouisLeNezet:
Is it better ?

PR #1629, opened 2025-06-04
Add more single cell test data

nictru:
This data will be the base for testing a large number of single-cell quality control modules that I will add soon


LeonHafner (approved):
LGTM

PR #1642, opened 2025-06-11
MAG: Rename test_data/minigut_cat.tar.gz to databases/cat/minigut_cat.tar.gz

jfy133:
Moving it for consistence with GTDBTK and BUSCO

Has not yet been used in earlier versions of the pipeline, so fine to move curently


erikrikarddaniel (approved):
:+1:

PR #1667, opened 2025-06-25
add yte config

famosab:


famosab:
For: https://github.com/nf-core/modules/pull/5886

jasmezz (approved):
LGTM. I would prefix both config files with the tool name, but I see that not all config files in the test-datasets folder do that. So no strong opinion from my side.

famosab:
The .yte is neccessary for the tool :) And I did not want to be too redundant.

PR #1683, opened 2025-07-17
Add .gff files from pyrodigal as the pre-annotated ones

HaidYi:
To use pre-annotated `.gff` files in the cazyme screening testing, manually generated them from `sample_1.fasta.gz` and `sample_2.fasta.gz`.


jfy133:
@HaidYi looks good, could you just write a sentence in the README how you generated them exactly (e.g. pipeline/tool/version/command etc.) ?

Npte: we haven't got that for all the files in this branch because we only added it as an internal policy recently... but you should see a few examples

jfy133 (approved):
Perfect thank you @HaidYi !

PR #1685, opened 2025-07-21
Updating datasets for bascule, mobster, and adding more data to viber

elena-buscaroli:
Hi! I've updated the input for the bascule module (providing a tar.gz folder instead of single files) and for the .csv input file for the mobster module (I've updated the module as well and I've added a column that was required). I've also uploaded an additional .csv dataset to test the viber module.


jfy133 (approved):

elena-buscaroli:
Thanks!

PR #1695, opened 2025-07-30
Add pfam small

rannick:


fellen31 (commented):
Could you add an entry to the README?

PR #1700, opened 2025-08-06
Fastq files without /1 /2 in names, umi in readnames bam

SPPearce:
Most (all?) of the current human fastq files have been generated by extracting reads from an already aligned bam file.
This leaves the read headers looking like:
`@922332/1`
for R1 and 
`@922332/2`
for the corresponding R2.
This causes failures when you try and extract umis using fastp, as the umi gets put after the `/{1,2}`, and therefore the names are not the same.

I'm also adding a human bam file where the umi has been put into the read header, suitable for tools that expect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,7 +445,7 @@ The earth sciences folder contain subfolders for different data formats encounte
         - test.paired_end.duplex_umi_mapped_tagged.bam: file originating from `test.paired_end.duplex_umi_unmapped.bam` and `test.paired_end.duplex_umi_mapped.bam`
         - test.paired_end.duplex_umi_grouped.bam: file originating from `test.paired_end.duplex_umi_mapped_tagged.bam`
         - test.paired_end.duplex_umi_duplex_consensus.bam: file originating from `test.paired_end.duplex_umi_grouped.bam`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -445,6 +445,7 @@ The earth sciences folder contain subfolders for different data formats encounte
         - test.paired_end.duplex_umi_mapped_tagged.bam: file originating from `test.paired_end.duplex_umi_unmapped.bam` and `test.paired_end.duplex_umi_mapped.bam`
         - test.paired_end.duplex_umi_grouped.bam: file originating from `test.paired_end.duplex_umi_mapped_tagged.bam`
         - test.paired_end.duplex_umi_duplex_consensus.bam: file originating from `test.paired_end.duplex_umi_grouped.bam`
+        - test.paired_end.umi_in_header.bam: Modification of `test.paired_end.unsorted_tagged.bam` to move the UMIs to the read header, and swap chromosome to entire chr22 rather than chr22:16570000-16610000
```
jfy133, line 448: Missing bai?

jfy133 (approved):

README.md
SPPearce: @jfy133 , added a mention of the index.

PR #1709, opened 2025-08-19
Add taxdump.tar.gz

jcbioinformatics:
Add taxdump.tar.gz file with just information relevant to tax ids of the SARS-CoV2 and _Haemophilus influenzae_ taxa under `metagenome/fasta/`

Needed for nf-core/modules PR https://github.com/nf-core/modules/pull/8680


jfy133 (approved):
Could maybe in the README list the files contained but not mandatory

jcbioinformatics:
I realized while updating the README that I didn't do a great job of subsetting the divisions and gencode files, so that's now fixed.

Are my descriptions in the README alright as far as like length and level of detail go?

jfy133 (approved):
Yup looks good to me! 

I had noticed also that first time around but assumed the two files weren't that important 😬

Before you merge just double check the indentation of the `.dmp`/`*.txt` list is in the right place - I think it needs to go one ident further

jfy133:
Thanks @maxulysse !

you're good to merge @jcbioinformatics !

PR #1718, opened 2025-08-26
Add modern amplicon sequencing testdata

BioWilko:
All of the current data is poorly labelled (or not at all) and is using ancient schemes / BED file standards, this adds data which is actually useful for testing amplicon workflows!


famosab:
Can you add the files to the README as well? :)

BioWilko:
> Can you add the files to the README as well? :)

Added

PR #1748, opened 2025-09-25
add data for modkit/repair

jkh00:
Added a small set of Oxford Nanopore reads from Arabidopsis thaliana (ecotype: Col-0), chromosome 1.
The two sets of reads represent ONT reads before trimming and after trimming. The trimming was done with `Porechop`. The purpose of having these two datasets is that `modkit repair` can correct the MM/ML tags on the trimmed reads using the untrimmed reads.


famosab:
Can you add it to the README as well? :)

PR #1754, opened 2025-10-04
Add v4 compatible AMRFinderplus output

jfy133:
The latest version of AMRfinderplus (v4) changed it's output file structure slightly.

Uploading a compatible file taken from teh AMRfinderplus website: https://github.com/ncbi/amr/blob/4d9d45dc2ac0f06b004623ecec1fa46aee752431/test_prot.expected

So that we can correctly test `hamronization/amrfinderplus`


jfy133:
It's in delete me thus no README

sateeshperi (approved):

PR #1767, opened 2025-10-29
add gold-standard reeference genome for SARS-CoV-2

Krannich479:
This PR adds the _de facto_ gold-standard SARS-CoV-2 reference (MN908947.3 / Wuhan-Hu-1) to the nf-core test datasets.

I need this reference to test the [upcoming nf-core module](https://github.com/nf-core/modules/issues/9231) of the [devider](https://github.com/bluenote-1577/devider) metagenome assembler.

It is hard to believe for me that this reference has never been uploaded yet. However, I scanned the `modules` branch as well as the `viralrecon` pipeline branch and to the best of my knowledge this reference is nowhere to be found. FYI, the SARS-CoV-2 (SC2) [genome ](https://github.com/nf-core/test-datasets/blob/modules/data/genomics/sarscov2/genome/genome.fasta) present in the modules branch is (to me) a random SC2 haplotype that I cannot use for metagenome assembly since this haplotype might shares mutations with other test data where I'd expect a different allele. In contrast, Wuhan-Hu-1 differs from the Beta and Delta SC2 variants of concern that I intend to use for metagenome assembly and haplotagging.

The sequence record in the FASTA file was downloaded from [ENA](https://www.ebi.ac.uk/ena/browser/api/fasta/MN908947.3) with a simplified header.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,8 @@ The earth sciences folder contain subfolders for different data formats encounte
     - 'genome.fasta': MT192765.1 genome including (GCA_011545545.1_ASM1154554v1)
     - 'genome.fasta.gz': bgzipped version of 'genome.fasta'
     - 'genome.fasta.fai': fasta index for 'genome.fasta'
+    - 'genome.MN908947.3.fasta': The gold-standard MN908947.3 genome (Wuhan-Hu-1) of SARS-CoV-2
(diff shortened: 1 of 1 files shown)


Krannich479:
I will close this as I was pointed to the following repo:
https://github.com/nf-core/test-datasets/tree/viralrecon/genome/MN908947.3

PR #1793, opened 2025-11-21
Add complete EV taxid list

AnnaNoren:
Addition of complete enterovirus taxid list.
Update of README

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,9 +113,9 @@ List of taxids used for testing filtering of blast database using `-taxidlist` o
 
 #### `blastdb/complete_ev_taxidlist.txt`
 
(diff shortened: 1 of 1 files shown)


genome/blastdb/complete_ev_taxidlist.txt
maxulysse: how is this list sorted?

AnnaNoren: @maxulysse Not completely sure what you mean but it is downloaded straight from ncbi and ranges in descending magnitude of "number", so enterovirus parent taxid 12059 is at the end of the list. The sorting does not affect the usage in blastn.

PR #1798, opened 2025-11-26
Add more CSV examples

dialvarezs:
With slightly different structure to the current one, so we can test CSV "smart" concat (https://github.com/nf-core/modules/pull/9459)


jfy133 (approved):
Double check you don't need to add entries to the main readme describing the files but otherwise looks good

PR #1809, opened 2025-12-01
test data for plastid psite and make_wiggle

suhrig:
This PR adds test data for the new submodules `psite` and `make_wiggle`.

Let me know if this is not the right place to put the test data.

The related modules have been pushed to here: https://github.com/nf-core/modules/pull/9482


pinin4fjords (changes requested):
There's a readme at the top level you need to add these files too. It would also be good if you added a README in this folder describing how the files were produced in case future-us needs to know.

suhrig:
I have added a description to the README files. Let me know if you would like any additional info.

pinin4fjords (approved):

PR #1825, opened 2026-01-06
Same sample ID

FernandoDuarteF:
The last two rows of the `samplesheet_test_illumina_amplicon.csv` share the same sample ID, which results in them being grouped together.

I may be mistaken, but I don’t think this is correct.

It would be nice to have a review in case it breaks any tests.

### DIFF
--- a/samplesheet/samplesheet_test_illumina_amplicon.csv
+++ b/samplesheet/samplesheet_test_illumina_amplicon.csv
@@ -2,4 +2,4 @@ sample,fastq_1,fastq_2
 SAMPLE1_PE,https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/illumina/amplicon/sample1_R1.fastq.gz,https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/illumina/amplicon/sample1_R2.fastq.gz
 SAMPLE2_PE,https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/illumina/amplicon/sample2_R1.fastq.gz,https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/illumina/amplicon/sample2_R2.fastq.gz
 SAMPLE3_SE,https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/illumina/amplicon/sample1_R1.fastq.gz,
-SAMPLE3_SE,https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/illumina/amplicon/sample2_R1.fastq.gz,
+SAMPLE4_SE,https://raw.githubusercontent.com/nf-core/test-datasets/viralrecon/illumina/amplicon/sample2_R1.fastq.gz,


mashehu:
oops, this broke our tools CI tests that rely on this samplesheet. but fixed now so no worries 🙂

svarona:
This is a test to check whetheir the are merged or not! Thay should be kept with the same ID!

jfy133:
> oops, this broke our tools CI tests that rely on this samplesheet. but fixed now so no worries 🙂

That suggests to me tools needs to use tests from a pipeline not under active development ;)

jfy133:
Also @chriswyatt1 please at least leave a comment saying you approve, don't just merge straight in!

mashehu:
nah, this is a good test for test data stability 😉

jfy133:
> nah, this is a good test for test data stability 😉

Touche!

chriswyatt1:
Sorry about that, we had wrote a detailed message, not sure what happened there. Maybe it didn't send. We tested a few pipelines to ensure it wasnt being used by another pipeline or break anything. But seemed like a mistake so merged it

jfy133:
> Sorry about that, we had wrote a detailed message, not sure what happened there. Maybe it didn't send. We tested a few pipelines to ensure it wasnt being used by another pipeline or break anything. But seemed like a mistake so merged it
>  

Ok! I said that because that information is indeed useful (and to me would imply this should be supported once tools doesn't break ;))

But also as general rule - when it comes to sample sheets it's best not change them once included in a release: if you change it it may break older versions of pipelines (which may still need to be supported for reproducibility purposes).

I would recommend just making a new file and update the config for the new pipeline release

FernandoDuarteF:
Sorry for opening this PR! I genuinely thought this was supposed to be a valid test since it's part of the defult test profile in the nf-core template.
But I now realise should have opened an issue first or mentioned it in slack.

jfy133:
> Sorry for opening this PR! I genuinely thought this was supposed to be a valid test since it's part of the defult test profile in the nf-core template.
> But I now realise should have opened an issue first or mentioned it in slack.

No need to apologise for opening a PR! It was more the merge seemingly before the review (which has the hiccup) that was the issue here :).

viralrecon is an important pipeline used in many different places so we need to give it an extra bit of care :)

svarona:
> Sorry for opening this PR! I genuinely thought this was supposed to be a valid test since it's part of the defult test profile in the nf-core template. But I now realise should have opened an issue first or mentioned it in slack.

As James said, it's perfect to open a PR and you did suggest me as reviewer, I would have explained the use of the same sample name and everything solved!

But changeing the sample name breaks all the pipeline unitary test, as it will generate another sample in the results and prior to the hackathon it would have break every new PR hahahaha

But you're very wellcome to open PRs to fix anyproblem you notice in the pipeline! We actually need more help with this pipeline!

mashehu:
@svarona maybe add @nf-core/viralrecon as a codowner for the whole repo in your branch, similar to #1888

PR #1834, opened 2026-01-15
Test data for Gens input data module

Jakob37:
Minimal test data for the WIP nf-core module:

https://github.com/nf-core/modules/pull/9656


FriederikeHanssen (changes requested):
could you add a note about the data here: https://github.com/nf-core/test-datasets/tree/modules/data/genomics/homo_sapiens/README.md? It's an imperfect system but at least we have some note on where the data came from

Jakob37:
> could you add a note about the data here: https://github.com/nf-core/test-datasets/tree/modules/data/genomics/homo_sapiens/README.md? It's an imperfect system but at least we have some note on where the data came from

Yes, makes sense.

I added a paragraph about how the data has been generated. It is not fully reproducible, but should at least give some context.

I am unsure about the guidelines here. Let me know if I am still missing something 🙏

FriederikeHanssen (approved):

PR #1851, opened 2026-01-28
Adding toy samplesheet (synthetic data)

quentinblampey:
This is a dummy samplesheet. Under the hood, Sopa will create a synthetic toy dataset to run the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,38 +1,8 @@
 # ![nfcore/test-datasets](docs/images/test-datasets_logo.png)
 Test data to be used for automated testing with the nf-core pipelines
 
-> ⚠️ **Do not merge your test data to `master`! Each pipeline has a dedicated branch (and a special one for modules)**
+## Test data and samplesheets for Sopa
 
-## Introduction
+Some samplesheets are available under the `samplesheet` directory:
 
-nf-core is a collection of high quality Nextflow pipelines. This repository contains various files for CI and unit testing of nf-core pipelines and infrastructure.
-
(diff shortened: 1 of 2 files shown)


mashehu:
please add a more detailed description about that in the readme

quentinblampey:
Thanks @mashehu, I updated it!

README.md
```diff
@@ -1,38 +1,8 @@
...
-git remote set-branches --add origin [remote-branch]
-git fetch
-```
-
-## Support
-
-For further information or help, don't hesitate to get in touch on our [Slack organisation](https://nf-co.re/join/slack) (a tool for instant messaging).
-
-[^1]: From [stackoverflow](https://stackoverflow.com/a/60846265/11502856)
+- `toy.csv`: a toy samplesheet (under the hood, Sopa will create a synthetic dataset).
```
mashehu: do you have some documentation on what Sopa is doing to generate this dataset?

quentinblampey: I added a short description and also a link to the Sopa API for this toy dataset (which includes a docstring and also the source code)

mashehu:
Sorry to be a pain, but why not just put the toy dataset directly in this branch instead of going through a generation step?

quentinblampey:
For multiple reasons:
1. The generation is very quick, so it's easier to run it rather than reading from a static file
2. The input of Sopa is raw data that we process into a SpatialData object (via a process called `TO_SPATIALDATA`). There is no "raw data" for the toy dataset since it's synthetic (but we still call `TO_SPATIALDATA` with a specific technology argument). Instead, if I directly save the toy dataset as a SpatialData object, this would skip the `TO_SPATIALDATA` process, and I think it's better to keep all processes in the tests.

In brief, doing this would be both more complex and would remove some parts of the process being tested.

mashehu (approved):
okay, makes sense. 🙂

PR #1861, opened 2026-02-09
add test files for hifitrimmer filterbam

sainsachiko:
Add `test.fasta.gz` and `test.fastq.gz ` for input testing of hifitrimmer_filterbam


muffato (changes requested):
The files look OK, but don't put them under `delete_me`

prototaxites:
I would say that delete_me is the correct location for these, as they are files that are directly intended to be run for a specific tool and should be removed if/when a generic replacement is found?

muffato:
What "generic replacement" are you envisaging ? Do you mean the big overhaul of the this repo ?

prototaxites:
To quote the `delete_me` documentation:

> The delete_me folder does not adhere to a defined structure as data in this folder should be delete as fast as possible, whenever a more suitable dataset is found that can be added to any other folder.

The idea is that tests should use data in the standard locations, if there is suitable data in there. You can add more data to these locations, but it needs to follow the structure and be well-documented. If you need a test file for a specific tool, it should instead go in `delete_me` until such a time as it can be replaced by a file that will fit within the standard data location/structure that can be used by many tools.

muffato:
In this case, I propose the to move the data to `data/genomics/eukaryotes/Pipistrellus_nathusii/genome/`

sainsachiko:
> In this case, I propose the to move the data to `data/genomics/eukaryotes/Pipistrellus_nathusii/genome/`

I have updated location for genomic data (BAM, FASTQ, FASTA). Location of `.bed.gz`; `.yaml`; `blastout.gz` remains unchanged as these are specific to `hifitrimmer`.

muffato (approved):

PR #1863, opened 2026-02-09
Add rgi card-data file

vinisalazar:
Hi,

This PR also creates a `metagenomics` directory. Although there is already a `metagenome` directory under `data/genomics/prokaryotes/metagenome/`, there is a discussion in the `#modules` nf-core Slack channel to add more files for metagenomics testing.

I propose that `metagenomics` should be a directory under the `data/` directory, at the same level as `genomics` and `metabarcoding`. The `prokaryotes` directory where it currently sits contains mostly WGS data for model organisms, which is a different scope. This PR introduces this change. I would like to follow up with additional PRs to move content from the `data/genomics/prokaryotes/metagenome` to this new directory, fix the corresponding tests, and add more files for metagenomic testing.

cc @jfy133 @prototaxites @SPPearce 

Best,
Vini

---

**Summary of changes:**
- Add data for testing the rgi modules. See nf-core/modules#9873 for discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,11 @@ The genomics folder contains subfolders for all organisms for which test data is
 Additionally there is a special subfolder for metagenome related files
 
 - metagenome
+  - fasta
+  - rgi
+  - taxonomy
+
+The `fasta` directory contains metagenomic sequence data, the `rgi` folder contains Resistance Gene Identifier data, and the `taxonomy` directory contains metadata for taxonomic classifier (TaxIDs, etc).
 
 All folders are structured in a similar way, with any genome-specific files in `genome` (e.g. fasta, gtf, ...) and technology specific raw-data files in the `10xgenomics`, `illumina`, `nanopore`, `pacbio`, `hic` and `cooler` subfolders whenever available.
 `Genomics` contains all typical data required for genomics modules, such as fasta, fastq and bam files. Every folder in `genomics` corresponds to a single organism. For every data file, a short description about how this file was generated is available either in this description or in the respective subfolder.
--- a/README.md
+++ b/README.md
@@ -40,7 +40,6 @@ Additionally there is a special subfolder for metagenome related files
   - rgi
   - taxonomy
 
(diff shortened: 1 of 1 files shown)


SPPearce:
We don't really like moving data that is already in the test-datasets, but adding it here sounds sensible.
Can you update the README.md, that has a list of the data sources of the files (although it isn't always up-to-date)...

jfy133:
Hrm, I see the logic @vinisalazar but I also agree with @SPPearce that splitting the source material for all of `metagenomics` is not a good idea, it will make discoverability more difficult.

I would rather keep everything in one place, and ignore the structure of `data/` I feel actually this has been somewhat abused and people have started adding things they should not (personal opinion ;) .

For context: I felt the logic was:

- genomics: DNA
- proteomics: proteins
- imaging: microscopy/images
- earth_sciences: special stuff

but people have started adding top level metabarcoding, and pangenomics which are both DNA so IMO _should_ fall under 'genomics' but unfortunately are not.

I would personally rather you stick to the original scheme and just live with the 'pain' of the others...

What do you think otehrwise @SPPearce ?

And I agree iwth the README update, regardless where you pu the  files eventually

SPPearce:
At some point we are planning to change to a completely different structure, but that still hasnt happened

vinisalazar:
> At some point we are planning to change to a completely different structure, but that still hasnt happened

That seems like a good objective for a future sprint! :)

SPPearce (commented):
Can you also add details of the file you are adding into the messy section below.

jfy133:
Yeah, documentation of their source, how they were generated would be helpful :)

vinisalazar:
Added a link, and the tar file also contains a README inside

README.md
```diff
@@ -36,6 +36,11 @@ The genomics folder contains subfolders for all organisms for which test data is
 Additionally there is a special subfolder for metagenome related files
 
 - metagenome
+  - fasta
+  - rgi
+  - taxonomy
+
+The `fasta` directory contains metagenomic sequence data, the `rgi` folder contains Resistance Gene Identifier data from the CARD database (May 2025 release, see https://card.mcmaster.ca/download for details), and the `taxonomy` directory contains metadata for taxonomic classifier (TaxIDs, etc).
```
SPPearce: I think this wants to be below, around line ~290

SPPearce (approved):

PR #1868, opened 2026-02-12
 Test data for nf-core/Deepmutscan

BenjaminWehnert1008:
As discussed in the Slack channel (#deepmutscan) some while ago, the deepmutscan pipeline requires unique test data, which is herewith ready to be uploaded and used.


.DS_Store
mashehu, line 1: please untrack the .DS_STORE files (there are two in this PR)

mashehu (changes requested):
Please remove the `.DS_STORE` files

Usual question: Do you need all these replicates for a minimal test setup?

.DS_Store
BenjaminWehnert1008: oops, my bad. I removed the .DS_STORE files.

For the replicates: We had fewer replicates implemented first for testing and changed it later. @MaximilianStammnitz, do you think we could remove one or even two replicates without losing functionality? I guess for DiMSum we should at least keep two replicates, right?

MaximilianStammnitz: > oops, my bad. I removed the .DS_STORE files.
> 
> For the replicates: We had fewer replicates implemented first for testing and changed it later. @MaximilianStammnitz, do you think we could remove one or even two replicates without losing functionality? I guess for DiMSum we should at least keep two replicates, right?

@BenjaminWehnert1008 @mashehu Yes, in the smallest test version I would recommend that we stick to a minimum of two input + two output replicates, just so that DiMSum (and other fitness estimation tools in the downstream module) can run through all of the error modelling steps.

MaximilianStammnitz: (processing these 4 x 50k PE reads should only takes 2-3 minutes)

mashehu: Then let's go with two replicates. We have the AWS megatests for more representative runs.

BenjaminWehnert1008: @mashehu @MaximilianStammnitz done:)

MaximilianStammnitz:
great, @mashehu ready to merge?

mashehu (approved):
a lot better thanks!

BenjaminWehnert1008:
@mashehu, anything we can do to complete the merge?

PR #1871, opened 2026-02-16
 added test data files for pharmcat module

ramsainanduri:
This PR adds minimal test data files required specifically for the PharmCAT module.

These files are not present in any existing nf-core test datasets because they are highly specific to PharmCAT input requirements. The test data has been intentionally kept very small to comply with nf-core testing guidelines and to ensure fast CI execution.

## Details

The following files are added:
- `test.single.sample.vcf.gz` – Preprocessed PharmCAT positions VCF in PharmCAT-specific format
- `test2.single.sample.vcf.gz` – Same as above, including non-reference alleles
- `test.outsideCall.tsv` – Minimal haplotype file formatted according to PharmCAT specifications
- `test.chr22.outsideCall.tsv`: Tab-separated file containing haplotype information derived from external calls for chr 22.
- `test.chr22.single.sample.vcf.gz`: Output generated from the PharmCAT positions VCF for chr 22 after preprocessing into the PharmCAT-specific format.


ryanjameskennedy (approved):
LGTM!

PR #1906, opened 2026-03-12
Include npmsfam db

npechl:
Partially fixes https://github.com/nf-core/proteinannotator/issues/77.


vagkaratzas (changes requested):
Name should be `nmpfams`  not `npmsfam` xD... the lines seem to have been pushed a bit inside: `//` should be at the start of the line as in the initial file.

Did you keep two models that have hits to the test input sequences?

npechl:
> Name should be `nmpfams` not `npmsfam` xD... the lines seem to have been pushed a bit inside: `//` should be at the start of the line as in the initial file.
> 
> Did you keep two models that have hits to the test input sequences?

Should be clear now.

vagkaratzas:
Nice, go for it!

PR #1917, opened 2026-03-12
Update models for clair3

EfraMP:
I'm updating the clair3 module. A major version of the tool has been released (v1.2 to v2.0), switching from TensorFlow to PyTorch.

An update of the models is needed in order to make it work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1020,7 +1020,8 @@ The earth sciences folder contain subfolders for different data formats encounte
   - 'rrna_reads_1.fastq.gz': Synthetic 5.8S rRNA paired-end reads (R1), 10 reads from rfam-5.8s-database-id98 for testing rRNA removal tools
   - 'rrna_reads_2.fastq.gz': Synthetic 5.8S rRNA paired-end reads (R2), matching R1 file
 - models
-  - clair3_hifi.tar.gz: gzipped tar folder containing models needed to run clair3 on Hifi data
+  - clair3_hifi.tar.gz: gzipped tar folder containing TensorFlow models needed to run clair3 on Hifi data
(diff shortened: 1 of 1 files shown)


jfy133 (commented):
I think it might be safer to just upload a new version so older versions of the module still works, rather than modify the existing file (what do you think @nf-core/maintainers ?)

SPPearce:
Yes, I would agree, I don't like overwriting files here.

vagkaratzas:
Separate file for now; may delete old / unused in the future

EfraMP:
Cool, I changed the name of the new file and updated the readme

pinin4fjords:
I would argue that we should let git handle the versioning rather than creating new files. Anyone referring to a file in Git and expecting a static file without using a SHA is a very silly person.

EfraMP:
Happy to go either way.
Please lmk which approach you'd prefer and I'll update accordingly.

maxulysse:
I'd agree with @pinin4fjords, but in the case of the test datasets for modules since we always use the modules branch, we do need to version in a way so that we don't modify old files and only add new ones.
So this looks ok enough to me.

maxulysse (approved):

jfy133:
> I would argue that we should let git handle the versioning rather than creating new files. Anyone referring to a file in Git and expecting a static file without using a SHA is a very silly person.

We should, but it's not how it's being used sadly. This would require a massive change and would break a lot of retroactive things - so I think this ship has sailed until we overhall test-datasets entirely

pinin4fjords:
Apologies, didn't think that comment through!

PR #1956, opened 2026-03-27
added testdata for anarcii

Clara0611:
Short fasta file with one TCR chain amino acid sequence per entry to be used as testdata for anarcii (PR #11058, currently in draft)

Stitchr testdata was run through stitchr and transformed into a fasta file.


miraep8 (approved):
Looks good to me! :) Nice test data Could be nice to include the exact command you used to do the subsetting.

Clara0611:
Thanks for reviewing! :) You're right, my wording was vague. Tbh I just copy and pasted the first few lines into a new file. I changed my wording to hopefully make this clear and added some detail on the fasta transformation as well.

PR #1960, opened 2026-03-30
Adding percolator test files

julianu:
Adding test files for Percolator module (PR opens after this is one is merged for the tests).

Until now, there are only RAW files and FASTA databases, which makes any tests for Percolator also run the conversion of the raw files and the mass spectra identification. This will make tests much easier.


jonasscheid:
@julianu Could you take the .pin output from your recent Comet nf-core module and just use that to test percolator? That way, we would not need the pin test datasets. See [here](https://github.com/nf-core/modules/blob/master/modules/nf-core/openmsthirdparty/cometadapter/tests/main.nf.test) how to use nf test with setting up input files. Does that make sense?

julianu:
Yes, I could use the Comet output directly. But is this not running Comet before the actual Percolator, or just using a (cached) output of it? I just thought it better to run as few as possible before the actual intended test. But when this is common practice, I could definitely do so.

jonasscheid:
>But is this not running Comet before the actual Percolator, or just using a (cached) output of it? 

Yes it would run Comet before this (which might take a bit of time), but the common goal is to keep the testdata repo as small as possible, therefore "building" testdata during CI run > providing testdata for each nf-core module.

The `setup` section in the nf-test lets you do this (see my link)

julianu:
All right, thanks. Will implement it like you suggest.

PR #1961, opened 2026-03-31
Adding test data for Comet

julianu:
adding comet params and an mzML conversion of Thermo RAW file for protein identification

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Additionally there is a special subfolder for metagenome related files
   - rgi
   - taxonomy
 
+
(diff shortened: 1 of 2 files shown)


jfy133 (commented):
Overall OK, but ore information on the source/how each file was generated would be helpful.

julianu:
@jfy133 I added some more information to the README. Is this sufficient, or more needed? 
The sage-file actually was already in the proteomisc path, but no info in the README .

jfy133 (approved):

jfy133:
> @jfy133 I added some more information to the README. Is this sufficient, or more needed? 
> The sage-file actually was already in the proteomisc path, but no info in the README .

- Ok if you didn't add the file don't eorry

Looks better now - do you have any info on the sample source the data was generated from? But I recognise these are extra files

julianu:
I added the information for the files I could find. Somehow they ended up to become default test files in the last years, as they are so small.

PR #1965, opened 2026-03-31
Adjusting Comet testdata

julianu:
The test data for comet had default settings, which I wanted to set in the Comet module.
This approach though breaks some nf-core defaults, so I changed the parameter file to be specific to the provided spectra file and adjusted the (upcomoing) module. No larger breaks there, only moving the logic slightly.


maxulysse (approved):
OK for me to change the file as it was not used yet.
cf https://github.com/nf-core/modules/pull/11096

PR #1970, opened 2026-04-07
Add test data for portello module

sofiademmou:
Add new test data for the portello module. 
The setup for this module/tool is quite specific so it needed its own small dataset.


peterpru (approved):
Looks good to me!

PR #1973, opened 2026-04-08
Centrifuger test data, .cfr files

haris18s:
Centrifuger database index files to be used for downstream classification.


mashehu (approved):

mashehu:
There is no possibility for moving it into one of the more specific directories, e.g. genomics?

haris18s:
In a similar case (centrifuge tool), authors (@sofstam), have placed it under delete_me. She kindly suggested to me, to do the same here as well.

haris18s:
> There is no possibility for moving it into one of the more specific directories, e.g. genomics?

I dont see where in genomics can fit, maybe in generic or if existed a metagenomics dir? OR any other suggestion :)

mashehu:
sorry, forgot to reply. your explanation about preceding cases was enough to convince me that `delete_me` is a good place for this.

PR #1981, opened 2026-04-10
Add hg38 cytoBand for telomerehunter module tests

pinin4fjords:
## Summary
- Add `data/genomics/homo_sapiens/genome/cytoBand_hg38.txt` from UCSC
- Needed by the telomerehunter nf-core module to pass the `-b` (banding file) flag

## Why
TelomereHunter defaults to a bundled hg19 cytoband when no `-b` flag is given. On hg38-aligned BAMs, 8 chromosomes are longer than the hg19 cytobands, causing an `IndexError` crash. The companion nf-core/modules PR adds a cytoband input to the telomerehunter module and needs this test data.

## Source
`http://hgdownload.cse.ucsc.edu/goldenpath/hg38/database/cytoBand.txt.gz`

1549 lines, tab-separated: chr, start, end, band, stain.


data/genomics/homo_sapiens/genome/cytoBand_hg38.txt
mashehu, line 1: we could probably just keep it to the standard chromosomes, but it's not so large as a file 
Is the standard extension for this file really `.txt`, not `.tsv`?

pinin4fjords: Good question - `.txt` is the canonical extension from UCSC (they distribute it as `cytoBand.txt.gz`). Tab-separated, but they don't use `.tsv`. I'd keep `.txt` to match the source and avoid confusion if someone cross-references with UCSC, but happy to rename if there's a convention in this repo.

On trimming to standard chromosomes - it's small enough to keep the full file (1549 lines, ~62 KB). Trimming could cause issues if someone has alt contigs or patches in their BAM.

mashehu (approved):

PR #1988, opened 2026-04-14
Metapep

xXFloBaerXx:
Add testdata for new protein input path


d4straub (commented):
Looks good, but would be nice to note the new dataset in the docu below https://github.com/nf-core/test-datasets/tree/metapep?tab=readme-ov-file#assembly-test-data?

d4straub (approved):
I approve already because I believe you will do whats necessary ;)

PR #1993, opened 2026-04-15
Rebuild centrifuger test dataset on native x86 -codespaces

haris18s:
Recreating the Centrifuger module test data after repeated segmentation faults. I suspect that  the previous index was built under Apple Silicon may be incompatible.
Thank you


data/delete_me/cfr_files_2.tar.gz
mashehu, line 1: could this be a smaller file?
Also how about replacing the older one? I don't think too many people used it yet, especially if it was broken

mashehu:
could you also drop the previous commit, so that it keeps the big file is not part of the git history. 
Do you think you need to mention in the Readme how you generated the not run into this in the future?

data/delete_me/cfr_files_2.tar.gz
haris18s: Ok, i created a smaller file, with a  smaller reference file. hope it works:) . 
Yes, we can delete i think

haris18s:
> could you also drop the previous commit, so that it keeps the big file is not part of the git history. Do you think you need to mention in the Readme how you generated the not run into this in the future?

Apologies, for late reply. I was running  the tests with the files under my  git - before pushing it here. 
-Passed the tests
-Added a README.md as well.

mashehu (approved):

PR #1994, opened 2026-04-16
add prunus persica input dataset

OlivierCoen:


mashehu (approved):
interesting choice of reference species 😁

OlivierCoen:
thanks for the rewiew!

PR #1999, opened 2026-04-17
Update GUNC test db to use official one

dialvarezs:


jfy133 (changes requested):
We should not delete the old one as it will break previous version of mag, just add the new test in addition :)

I would note in the original GUNC docs that that test file is no longer used, but kept for older versions accordingly

dialvarezs:
> We should not delete the old one as it will break previous version of mag, just add the new test in addition :)

Actually... the "old" mock db never worked, and I forgot to delete it.
I tried to enable GUNC with the mock db on this PR: https://github.com/nf-core/mag/pull/855

And we end up deciding to keep it disabled:
image
image

jfy133 (approved):
oh haha my bad here you go then!

PR #2002, opened 2026-04-20
Add unit test data for mivmir, gicam modules

torbjorgen:
Module test files for mivmir, gicam modules as annotated in rank variants subworkflow.


peterpru (approved):
Looks good to me!

PR #2016, opened 2026-04-27
Update database sheets to include the latest version of diamond

sofstam:


jfy133 (commented):
Could update the README to describe how yuou built the new DIAMOND databases (fine to just to say somehting 'like same as above but with DIAMOND vx.x.x')

jfy133 (approved):

PR #2017, opened 2026-04-28
Add reduced penetrance file for testing genmod models

fellen31:
Add reduced penetrance file for testing the genmod models module.


beatrizsavinhas (approved):
Looks good to me! 🌟

PR #2018, opened 2026-04-28
add samplesheet for full plate test

rboz1:


complete_plate_dataset/samplesheet_test_full.csv
mashehu, line 1: These it quite big, we usually prefer files <1mb if possible. Is even for a full test such a big sample sheet needed?

rboz1:
Yeah, agreed. I think this would take a long time to run just as a test
too. I could do ¼ of it? Then it’d be larger than the minimal test with 3
wells but not massive.

.DS_Store
kenibrewer, line 1: Let's delete this file, otherwise this is great.

kenibrewer (approved):
One minor change. Otherwise great.

PR #2022, opened 2026-04-29
Reduce genmod score config size/complexity

fellen31:
- Reduce genmod score config size/complexity by removing unnecessary criteria. 
- Update README to include the file, which was missing before.


beatrizsavinhas (approved):
Much cleaner! 🧹

PR #2025, opened 2026-04-29
feat: add test data for metator

prototaxites:
None of the existing data seems to work for the Metator module I'm developing, so porting their test data here so we can write a tested module.

https://github.com/nf-core/modules/pull/11370


sainsachiko (approved):
LGTM!

PR #2039, opened 2026-05-07
feat(gtdbtk): add new r232 gtdbtk database

prototaxites:
Adds the new test database for GTDBTk 2.7+, compatible with the new r232 release.


mashehu:
don't forget to update the readme as well

prototaxites:
Good call! Done.

databases/gtdbtk/gtdbtk_mockup_r232_20260507.tar.gz
mashehu, line 1: i am guessing there is no chance in making this even smaller?

prototaxites: This is the small version of the 56Gb database the tool requires, as provided by the developers. I think it's about as minimal as can be...

PR #2056, opened 2026-05-18
Add minimal FASTQ test data for create_reads_manifest module

timrozday-mgnify:
Two 20 bp single-read FASTQ files (fastq_1, fastq_2) used by the CREATE_READS_MANIFEST nf-test module test in seqsubmit.


timrozday-mgnify:
> Thank you! Did you run webin-cli validate for those reads? Does it work?

It does, on the ENA test service

PR #2059, opened 2026-05-19
Added sample list to test the cageronly mode

sidorov-si:


samplelist/cageseq_sample_list_test.csv
mashehu, line 1: I think the path might be wrong

sidorov-si: I put cageseq2_sample_list_test.csv in its own directory because it is for cageronly, not for the full pipeline run (and, I think, we usually call it sample list in the pipeline). This way, we will not get confused. So, the path to this file is correct (both the directory and the file exist).

mashehu: ah, okay, but maybe call it `cageseq_sample_list_test.csv` without the 2.

sidorov-si: Thank you Matthias! Done.

PR #2086, opened 2026-05-31
Add trio test data and update sample_id to sample

nourmahfel:


trio_testdata/trio.ped
mashehu, line 1: can you describe the origin of these test files in the readme file please

PR #2099, opened 2026-06-09
Updating test data

JudithBernett:
meta now includes DepMap sample info and tissue mapping, TOYv1 and TOYv2 now include drug_smilesvec and pathway_features for the newly added Precily model

TODO: 

- [ ] update documentation
- [ ] make data as small as possible


test_data/TOYv1.zip
mashehu, line 1: uff, that's quite a large file. any possibility to make it smaller, best case <1MB? This will make all CI tests faster (and keep this repo a bit slimmer).

JudithBernett: < 1MB is definitely impossible, but I think I can make some files smaller.
But the test data is already pretty minimal (sadly) - we need it to test all models, which can have so many different inputs in terms of omics (gene expression, methylation, proteomics, mutation, cnv, pathway features) and drug input (smiles, molgnet encodings, chemberta, graph, …). We also need the second toy data for testing cross-study generalization.

mashehu: looking into the file a bit... looks like `human_ppi_features.tsv` is the biggest file and looks to me to be sub-set-able.

JudithBernett: arrived at the same conclusion :)

mashehu: don't forget to drop the initial commit with the larger files via git rebase, to remove it from the git history.

JudithBernett: oh actually, they can become even smaller; will have to fix later

JudithBernett:
Update: turns out, we don't need this, we'll handle it in the testing :D will close the PR again

PR #2100, opened 2026-06-11
add reference gff3

piplus2:
This adds the gff3 version of the already available gtf to be used for testing dexseq modules.


piplus2:
it was already chrX... I just converted the gtf...

mashehu:
could you maybe switch to chr22 then?

piplus2:
Does it make any difference the fact that the authors subsampled chrX here?

mashehu:
ah, okay, then they should have subsampled it even more 😁 
Our goal for this repo is to have as small files as possible to make testing faster (and keeping this repo slim-ish). 
But if it is too much of an hassle, you can keep what you have.

piplus2:
I'm adding it to the todo list for the next release. Thanks for the feedback!

PR #2101, opened 2026-06-11
Funcprofiler fix path for for mifaser database

nickp60:
This fixes the path that the database tar unpacks to; previously it contained nested folders.


README.md
```diff
@@ -153,7 +153,9 @@ cd ../../../../
 mkdir data/database/mifaser
 git clone https://bitbucket.org/bromberglab/mifaser/src/3e41beae9746189e116d6ec6b15c73af68e9a99c/ tmpmifaser
 cp -r  tmpmifaser/mifaser/database/GS-24-all data/database/mifaser/
-tar czf data/database/mifaser/GS-24-all.tar.gz data/database/mifaser/GS-24-all/
+cd data/database/mifaser/
+tar czf GS-24-all.tar.gz GS-24-all/
+cd ../../../
```
piplus2: why don't you untar in the destination dir without cd so you can avoid the cd ../../../?

nickp60: done

PR #2107, opened 2026-06-17
Update test-data to work with latest version of pixelator (0.29.0)

elhb:
Updating test-data to work with the version update of pixelator (to 0.29.0) in this PR https://github.com/nf-core/pixelator/pull/227


Aratz:
Wouldn't this be the case for any panel update? Should we keep all previous versions of the test data and e.g. create one folder per version instead?

mashehu:
Exactly, that's how it is the best practice to not loose backwards compatibility.

Aratz:
Would it be ok to reference commits in `pipelines_testdata_base_path` (like nf-core/rnaseq does https://github.com/nf-core/rnaseq/blob/e7ca46272c8f9d5ceee3f71759f4ba551d3217a4/nextflow.config#L143)?

This way we don't risk breaking anything and we are quite free to update or reorganize the test data as we want.

mashehu:
yes, sounds like a good solution

elhb:
@mashehu I would like to suggest that we adopt the same strategy as [nf-core/rnaseq](https://github.com/nf-core/rnaseq/blob/e7ca46272c8f9d5ceee3f71759f4ba551d3217a4/nextflow.config#L143)
i.e. version test-datasets by specifying the commit sha in the pipeline repo.

By doing this we'd avoid creating new folders (with potentially near identical content) when some part of the testdata is updated.
We also achieve a clear connection of the pipeline version to specific test data version (the branch name pointer moves, while the commit sha is static).

Furthermore as the backwards compatibility likely already is broken by recent changes,
how about we adopt the commit sha strategy from here on i.e. we don't create additional folders now?

I already pointed the `nf-core/pixelator` PR to a commit as an example https://github.com/nf-core/pixelator/pull/227/changes/8a9e5e46286d27fbbf021aaa6523394db9bc3ecf (will ofc need to fix this to be the correct commit sha and repo once we decided how to move forward).

What do you think?

Thanks! //EB

elhb:
Haha, sorry hadn't update so didn't see @Aratz already suggested this :)

PR #2116, opened 2026-06-23
add misopy index for module testing

piplus2:
Added the misopy index generated using the gff3 in the reference folder.
This index will be used to test the misopy modules of the workflow.


nvnieuwk:
I'm guessing you will probably need to tarzip this, it won't be useful otherwise

piplus2:
these are all part of misopy index, the tool requires them to run. I don't know what's the best approach to use here. I can compress though

piplus2:
I also moved the tar.gz under reference since it's linked to the gff3

piplus2:
thanks!

PR #2117, opened 2026-06-24
add benchmarking data for samplesheet.tsv

d4straub:
I would like to add benchmarking datasets for CI tests for https://github.com/nf-core/ampliseq/pull/1021


d4straub:
Hm indeed the readme is just empty for ampliseq. It seems like that wasnt dont properly in the last half a decade...
In his case here the files are simply made up from expected results. I'll note down to revisit the README and try to find sources to the data files.

PR #2136, opened 2026-07-09
cellpheno/nis: add test tile + models

Chrisa142857:
Test data for the `cellpheno/nis` module (nf-core/modules#12179).

- `data/imaging/segmentation/cellpheno_nis/cellpheno_nis_tile.zip` — a real P4 partial-brain crop: 30 z-slices, 384×384 (2D OME-TIFF), ~4 MB.
- `data/imaging/segmentation/cellpheno_nis/cellpheno_nis_models.zip` — the TorchScript NIS models (cuda:0 set), ~64 MB.

Both are zipped so the module test extracts them via the `UNZIP` subworkflow (dir inputs), following the `numorph/3dunet` pattern. Verified: NIS segments 351 nuclei on this crop.


data/imaging/segmentation/cellpheno_nis/cellpheno_nis_models.zip
mashehu, line 1: this file is too big, we usually aim for as small as possible, target <1MB. Any chance you can shrink this and the second file?

Chrisa142857:
Thanks @mashehu! Addressed:
- **Shrank the tile** to a smaller real P4 crop (16 slices, 224×224) — now **~700 KB** (<1 MB).
- **Dropped the models zip** entirely: the TorchScript weights (~64 MB) are large and available via gdown from the source G-drive, so they don't belong in test-datasets — the module test fetches them at runtime instead of bundling them here.

So this PR now adds a single ~700 KB file. NIS still segments 221 nuclei on the crop.

PR #2139, opened 2026-07-10
hlatyping: real full-size data for samplesheet_full.csv

jonasscheid:
Replaces the `samplesheet_full.csv` placeholders (small HLA-fished FASTQs) with real full-size data, so nf-core/hlatyping's `test_full` can exercise the new FASTQ → GRCh38 → BAM path used by SpecHLA and HLA\*LA.

- `NA12878_WES` — Agilent WES, reused from nf-core/sarek's full test
- `GM12878_RNA` — RNA-seq, reused from nf-core/rnaseq's full test (same individual as NA12878, so DNA and RNA calls can be cross-checked)
- `HepG2_PEPTIDE` — peptides for immunotype; needs the new `tsv` column
- Drops `SAMPLE_BAM`: `example_pe.bam` has no `@SQ` dictionary, so HLA\*LA and SpecHLA cannot use it

Verified by running `-profile test_full` against this sheet. All tools reproduce NA12878's published genotype (`A*01:01`/`A*11:01`, `B*08:01`/`B*56:01`, `C*01:02`/`C*07:01`).

Needed by nf-core/hlatyping#222.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,16 @@ bwa mem -R '@RG\tID:NA11995\tSM:NA11995' \
 samtools index NA11995_GRCh38.bam
 ```
 
+### `samplesheets/samplesheet_full.csv`
+
+Full-size input for `-profile test_full`. NA12878 and GM12878 are the same individual, so DNA and RNA calls can be cross-checked.
+
+| Sample | Data | Source |
+| ------ | ---- | ------ |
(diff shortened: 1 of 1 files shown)


jonasscheid:
Documented the three files in the README under `## hlatyping test data`. Thanks!

PR #2163, opened 2026-07-22
Add gridss config file for gridss modules

imsarath:
## Description

 Adds a `gridss.properties` configuration file for the GRIDSS modules (`gridss/assemble`, `gridss/call`) to the test-datasets `modules` branch.

### Files added

   - `data/genomics/homo_sapiens/illumina/gridss/gridss.properties`
   - source: https://github.com/PapenfussLab/gridss/blob/master/src/main/resources/gridss.properties

### DIFF
--- a/README.md
+++ b/README.md
@@ -688,7 +688,7 @@ The earth sciences folder contain subfolders for different data formats encounte
       - test2.yak: Yak kmer index of 1000 of maternal reads from the GIAB Ashkenazim trio [RM8392](https://www-s.nist.gov/srmors/view_detail.cfm?srm=8392). These reads were selected from D3_S1_L001_R{1,2}\_001.fastq.gz and D3_S1_L001_R{1,2}\_001.fastq.gz so that they map to `pacbio/fastq/test_hifi.fastq.gz`.
     - gridss:
       - test.gridss.vcf: First 500 lines of a somatic structural variant vcf file generated using `gridss`, suitable for testing gridds/somaticfilter.
-      - gridss.properties: GRIDSS configuration properties file, suitable for testing the `gridss` modules. source: https://github.com/PapenfussLab/gridss/blob/825bc8e1bb092e9bb2ccb0c1c51dc6e7e1a922dd/src/main/resources/gridss.properties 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -688,6 +688,7 @@ The earth sciences folder contain subfolders for different data formats encounte
       - test2.yak: Yak kmer index of 1000 of maternal reads from the GIAB Ashkenazim trio [RM8392](https://www-s.nist.gov/srmors/view_detail.cfm?srm=8392). These reads were selected from D3_S1_L001_R{1,2}\_001.fastq.gz and D3_S1_L001_R{1,2}\_001.fastq.gz so that they map to `pacbio/fastq/test_hifi.fastq.gz`.
     - gridss:
       - test.gridss.vcf: First 500 lines of a somatic structural variant vcf file generated using `gridss`, suitable for testing gridds/somaticfilter.
+      - gridss.properties: GRIDSS configuration properties file, suitable for testing the `gridss` modules.
```
mashehu: where does the file come from?

imsarath: It's the default configuration file directly from the GRIDSS tool. I've also included the exact source file path in the PR description for reference.

imsarath: I realize now that it isn't strictly mandatory, but I included it as a reference to give other users a helpful example of the available options.

If you feel this file isn't necessary, just let me know, and I'm perfectly happy to close the PR. Sorry for the extra noise!

mashehu: so will this file be used as a test input somewhere?

imsarath: yes, I plan to use it as an optional configuration input in the nf-test for the gridss modules I'm working on (like gridss/assemble https://github.com/nf-core/modules/pull/12382 , gridss/call https://github.com/nf-core/modules/pull/12401) .

mashehu: then it's fine. pleaes add the source URL to the readme (and not just to the branch, but to the actual commit to pin it down as much as possible)

imsarath: Makes perfect sense! I have updated the README and added the permalink tied to the specific commit. Thanks for the review!

PR #2164, opened 2026-07-22
nf-core/ampliseq: Complete version 3 sample sheets

d4straub:
A recent PR in ampliseq required updating the major version to 3, see https://github.com/nf-core/ampliseq/pull/1027. Therefore, we decided also to adapt the sample sheets and make the standardized format required, see https://github.com/nf-core/ampliseq/pull/1028. So here are sample sheets for version 3 completed (some were planned earlier already).


d4straub:
Thanks!

PR #2177, opened 2026-07-24
added test data for mmusculus + clean up

JohannesKersting:


JohannesKersting:
@mashehu Thanks for your prompt comment and the notice. In fact, none of the datasets in the repo had documented origin. Added this information to the readme and also took the oportunity to remove some files which are no longer in use for testing.

JohannesKersting:
@mashehu Would you mind having a look again! :) Thanks in advance

mashehu:
Looking okay now. I guess we can't remove the deleted files from the git history without messing with the whole repo. But it is okay.

PR #2191, opened 2026-08-01
[funcprofiler] Make eggNOG-mapper data dir a tarball

vinisalazar:
emapper.py --data_dir takes a directory, but the database.csv row pointed at a bare eggnog.db file, so eggNOG-mapper failed with "invalid existing_dir value".


nickp60:
Thanks @vinisalazar!

PR #2196, opened 2026-08-04
Add riboseq UMI samplesheets

pinin4fjords:
## Summary

- add a mixed samplesheet with an explicit UMI override and a blank value that exercises the global fallback
- reuse the existing GSE182201 riboseq branch FASTQs

Required by nf-core/riboseq#223.


pinin4fjords:
Closing this because the scenario-specific samplesheet is generated directly in the nf-test. The test continues to use the existing riboseq branch FASTQ fixtures, so no new test-datasets file is needed.

PR #2207, opened 2026-08-07
Test bams for ont and pacbio

khersameesh24:
Added ont and pacbio bam files as test data for the oarfish module


kubranarci:
@khersameesh24 maybe add someone else from the core team for approve this (I am not sure about the procedure for test data in modules), Maxime might be on vacation.

PR #2217, opened 2026-08-17
Add minimal-dataset fixtures for singleton, align, and precalled-VCF test profiles in raredisease

ramprasadn:
## What's included

manifests/
- samplesheet_singleton.csv — proband alone, pointing at the existing fastq/ trio reads
- samplesheet_align.csv / samplesheet_align_singleton.csv — pre-aligned input, for test_align/test_align_singleton
- samplesheet_vcf.csv / samplesheet_vcf_singleton.csv — precalled-VCF input, for test_vcf/test_vcf_singleton

alignment/ 
- pre-aligned trio, for test_align's mixed-format samplesheet: two parents as BAM, the proband as CRAM (to exercise CRAM ingestion specifically), plus a BAM conversion of that same proband CRAM reused by the singleton profile (alignment is sample-level, so it's identical content either way — no separate singleton run was needed).

precalled/ 
- case-level precalled VCFs for case amusingmarmoset, one per variant type (snv/sv/mt/me/repeat), nested by scope:
- precalled/trio/ 
--  jointly-called across the family, proband-only samplesheet rows per pipeline convention (parents referenced via paternal_id/maternal_id)
- precalled/singleton/ 
--  same 5 types, called for the proband alone

README.md - documents all of the above.


sofiademmou (approved):
LGTM ! 🌟

PR #2218, opened 2026-08-18
raredisease: Update parent IDs in the precalled-VCF trio samplesheet

ramprasadn:
This PR updates the samplesheet for the VCF entrypoint for singletons.


fevac (approved):
LGTM! Please add a short description and note that this PR is still a draft

PR #2215, opened 2026-08-15
Add RPL28 eukaryotic test data for MetaEuk ORF calling (nf-core/metatdenovo #459)

erikrikarddaniel:
## Description

Adds a small eukaryotic test dataset for nf-core/metatdenovo's upcoming MetaEuk ORF-caller support ([nf-core/metatdenovo#459](https://github.com/nf-core/metatdenovo/issues/459)), and the two related consolidation issues it's designed to also cover ([#460](https://github.com/nf-core/metatdenovo/issues/460), [#463](https://github.com/nf-core/metatdenovo/issues/463)).

metatdenovo's existing test data has no eukaryotic genomic (intron-containing) content, so it can't exercise splice-aware gene calling or cross-caller/cross-contig ORF consolidation.

Source: the *Saccharomyces cerevisiae* S288C RPL28 locus (60S ribosomal protein uL15), real NCBI sequence, has a single well-annotated intron. Three read sets simulated from it with `wgsim` (genomic, spliced mRNA, and a synonymously-recoded variant of the same CDS) are engineered so that co-assembly with MEGAHIT produces two verified scenarios: a same-contig, splice-aware two-exon gene call, and a cross-contig, protein-identical pair of ORFs. Full construction details and exact assembly outcomes are in the branch's own `README.md` under "Datasets for nf-core/metatdenovo".

New files, all under `test_data/metaeuk/`:

- `rpl28_genomic_R1/R2.fastq.gz`
- `rpl28_mrna_R1/R2.fastq.gz`
- `rpl28_mrna_recoded_R1/R2.fastq.gz`
- `rpl28_reference_db.faa`

Plus `samplesheet/samplesheet_metaeuk.csv`, referencing the three read sets as three samples.

Already verified end-to-end against a real nf-core/metatdenovo pipeline run (`-profile test_metaeuk`, MetaEuk `--orf_caller`): assembles to 2 contigs as designed, MetaEuk calls the expected two-exon splice-aware gene, featureCounts sums both exon segments into one gene-level count.

Generated by Claude, reviewed by @erikrikarddaniel before opening.

PR #2219, opened 2026-08-18
Fix samplesheet_metaeuk.csv FASTQ URLs to point at nf-core/test-datasets

erikrikarddaniel:
## Description

Follow-up to #2215: the samplesheet added there left its `fastq_1`/`fastq_2` columns pointing at the fork (`erikrikarddaniel/test-datasets`) instead of `nf-core/test-datasets`, even though the same FASTQ files were uploaded to both. Every other samplesheet in this repo uses `nf-core/test-datasets` URLs; this brings `samplesheet_metaeuk.csv` in line.

Verified all 6 referenced FASTQ files resolve (HTTP 200) at their new `nf-core/test-datasets` URLs.


d4straub (approved):
LGTM!

PR #2223, opened 2026-08-18
Add dartseq reference genome/gtf and bullseye test fixtures; use full…

brovolia:
Add data/dartseq/reference/genome.fa and genes.gtf, the synthetic test genome referenced by the dartseq pipeline's test profiles.
Add data/dartseq/reference/bullseye.test.refFlat and data/dartseq/contrasts_test.csv, previously duplicated in the pipeline's own assets/ directory.
Rewrite samplesheet_4samples.csv to use full raw.githubusercontent.com URLs for fastq_1/fastq_2 instead of paths relative to the repo root, so the samplesheet resolves correctly regardless of working directory.


brovolia:
Closes #2223

erikrikarddaniel (approved):
I have been asked by Matthias to always update README.md so test files are documented. Approving anyway.

PR #2226, opened 2026-08-19
Add single-row diamond_dbs.csv variant for CI-lightweight multi-caller testing

erikrikarddaniel:
## Description

Follow-up to #2215/#2219 for [nf-core/metatdenovo#462](https://github.com/nf-core/metatdenovo/issues/462) (running multiple `--orf_caller` values in one execution).

The pipeline's multi-caller pipeline test needs `--diamond_dbs` on to catch a real class of bug (a Nextflow `.join()` silently dropping duplicate left-side keys when multiple ORF callers share the same diamond db name), but combining it with the full two-row `samplesheet/diamond_dbs.csv` pushed that CI job over the runner's disk budget (confirmed via repeated CI failures, "no space left on device").

`diamond_dbs_single.csv` is the same `ncbi-refseq-test` row already used in `diamond_dbs.csv` -- no new database files, just fewer rows -- as the only row, chosen since it also exercises the `parse_with_taxdump` code path. Still enough to reproduce the duplicate-key bug (two callers sharing one db name) at roughly half the diamond-side CI cost.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,4 +77,15 @@ test_data/metaeuk/rpl28_mrna_recoded_R2.fastq.gz
 test_data/metaeuk/rpl28_reference_db.faa
 ```
 
+### Single-row `diamond_dbs.csv` variant for CI-lightweight multi-caller testing
+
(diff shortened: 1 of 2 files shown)


danilodileo (approved):
LGTM

PR #2228, opened 2026-08-20
Updated samplesheet

FernandoDuarteF:
Added `fastq` files to `input_myco_tiny.csv` so that tests also run Merqury.


chriswyatt1 (approved):
fq back, great

PR #2229, opened 2026-08-20
feat(semibin2): add test data for semibin/multieasybin

prototaxites:
SemiBin2 in multi-binning mode requires the input sample FASTAs are concatenated in a specific way, and then the reads mapped per sample. This type of data is not available in the test-data so I've just ported a portion of their small test data.


erikrikarddaniel (approved):
👍

PR #2231, opened 2026-08-20
[modules] add test data for hyphy modules

leapicard:
In frame codon alignment of short protein (187 codons) and corresponding phylogenetic tree for evolutionary analyses testing.
Will be used for about 10 different modules.


erikrikarddaniel (approved):
Looks good. You have a prek failure, but it's not against your files I think. Did something else sneak in here?

leapicard:
> Looks good. You have a prek failure, but it's not against your files I think. Did something else sneak in here?

I'm not sure what could have?

PR #2232, opened 2026-08-24
Add hmmsearch table excerpts for hmmer/hmmrank module tests

erikrikarddaniel:
Adds `testdata/hmmrank/`: four gzipped HMMER 3.4 tables, a per-sequence (`--tblout`) and a per-domain (`--domtblout`) one for each of the `NrdJ` and `NrdJm1` ribonucleotide reductase profiles.

They are excerpts of a real `hmmsearch` of both profiles against proteins predicted from the GTDB r226 representative genomes, run through nf-core/phyloplace in "search and place" mode. Five target sequences were picked and every row for them kept verbatim, giving nine `(target, profile)` pairs. The only edit is that the trailing comment block `hmmsearch` writes at the end of each table has been removed, since it records the run's working directory and a timestamp.

The five targets cover the domain layouts the per-domain coordinate arithmetic has to handle: a single domain, several overlapping domains, several disjoint domains, and domains that overlap in profile coordinates while lying far apart on the target sequence. `AACBMDAH_04658` against `NrdJm1` is the last of those, spanning positions 14 to 803 on the sequence but covering only 400 of them, which is what a gene split over several open reading frames looks like in these tables.

Needed by an upcoming nf-core/modules PR adding alignment coordinates to `hmmer/hmmrank` (nf-core/phyloplace#70). The new dataset is documented in this branch's `README.md`.


CaroAMN (approved):
LGTM

PR #2240, opened 2026-08-26
Hyphy test data

leapicard:
Added shorter dataset (codon and aa msa + tree), and aa msa for existing test dataset.


Joon-Klaps:
Do you still need the other ones that got uploaded last week, if not I would delete.

leapicard:
> Do you still need the other ones that got uploaded last week, if not I would delete.

Yep still needed, they're used for different modules within the Hyphy ecosystem.

PR #2245, opened 2026-08-28
bppml test param file

leapicard:
Test data for bppml from bppsuite.


erikrikarddaniel (approved):
Reviewed with some help from Claude Code. Small plain-text file (120 lines), no size concern, and the README config section has already been updated with a description. Good to merge.

PR #2246, opened 2026-08-28
Updated field in samplesheet

FernandoDuarteF:
Changed `species` field to a more generic `assembly`.


chriswyatt1 (approved):
That makes sense. All fine

PR #2247, opened 2026-08-28
Add raw subread/featurecounts fixtures for custom/collectfeaturecounts

erikrikarddaniel:
## Description

Two small, per-sample `featureCounts.tsv` fixtures in the real `subread/featurecounts` raw output shape (`# Program:featureCounts` comment line + `Geneid/Chr/Start/End/Strand/Length/<bam>` header), each derived from a real run of `subread/featurecounts`'s own module test against `genomics/sarscov2` data, with per-sample count values varied.

For testing a new `custom/collectfeaturecounts` nf-core/modules component (mirrors `custom/collectstats`'s own test data added in #2210). Deliberately distinct from the existing `sample{1,2}.*.featureCounts.tsv` fixtures added in #2210, which are already-collected/joined tables (have `accno`/`tpm` columns) rather than raw tool output — the two shapes would otherwise collide in naming intent.

Generated by Claude

### DIFF
--- a/README.md
+++ b/README.md
@@ -1165,6 +1165,10 @@ The earth sciences folder contain subfolders for different data formats encounte
   - 'test.csv': exemplary comma-separated file obtained from [here](https://bioinf.shenwei.me/csvtk/usage/#split)
   - 'test2.csv': exemplary comma-separated file with user data (id, first_name, last_name, email)
   - 'test3.csv': exemplary comma-separated file with email and last_login timestamps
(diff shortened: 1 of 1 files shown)


leapicard:
Could you update the README?

erikrikarddaniel:
Thanks @leapicard . I took the opportunity to add descriptions to some quite recent other datasets I added.

leapicard (approved):

PR #2252, opened 2026-08-31
Store all chromosomes genetic map

LouisLeNezet:
To avoid depending on external github account, store now all chromosomes in our dataset branch.
Old genetic map for chr21 and 22 are deleted as already transfered to the modules branch.
These files are now only used in local sbwf `vcf_chunk_glimpse` and csv/map_glimpse.csv and csv/map_plink.csv they will be updated in a PR against phaseimpute/dev.


erikrikarddaniel (approved):
That's a lot of files and some of them large, but I assume you know what you're doing :-)

LouisLeNezet:
Thanks for the review.
The aim is to no longer rely on an external GitHub account for our unit tests.
We recently had an issue with the shapeit5 account closing and breaking our full test.

PR #2253, opened 2026-08-31
viralmetagenome: add SARS-CoV-2 UK annotation dataset (sequences + metadata table)

Joon-Klaps:
Adds 74 UK SARS-CoV-2 genomes downloaded from NCBI Virus together with the metadata table exported alongside them; the fasta headers carry no `key="value"` annotations of their own, so the two files have to travel as a pair. Needed to validate the new `--annotation_metadata` option in nf-core/viralmetagenome#318, which reads the annotation fields from such a table instead of parsing them out of the fasta headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,70 @@ git remote set-branches --add origin [remote-branch]
 git fetch
 ```
 
+## Data description
+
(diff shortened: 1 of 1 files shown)


erikrikarddaniel (approved):
I've been asked by others to keep a README.md describing the data. You do as you please.

PR #2254, opened 2026-08-31
phyloplace: grouped sample sheet for search-and-place mode

erikrikarddaniel:
## Description

Adds `testdata/phylosearch_input_reftree.csv` for [nf-core/phyloplace#78](https://github.com/nf-core/phyloplace/issues/78) — a two-row "search and place" sample sheet whose rows share a `reftreename` and are therefore summarised jointly as well as individually.

Follows on from #2250, which added the column to `phyloplace_input.csv`.

### Why a separate file

`phylosearch_input.csv` is deliberately left untouched, so it keeps covering the layout of a sample sheet written **before** the `reftreename` column existed. Between the two files the suite then tests both layouts in that mode: without the column, and with a group.

### What is in it

Two rows, both searching `testdata/domain_16s.fna` and placing onto the same archaeal 16S reference alignment and tree, one with barrnap's archaeal 16S profile and one with its bacterial one:

| target | profile | hits | `reftreename` |
| --- | --- | --- | --- |
| `arc16s` | barrnap `arc.hmm` 16S_rRNA | 11 | `arc16s` |
| `bac16s` | barrnap `bac.hmm` 16S_rRNA | 12 | `arc16s` |

The two profiles hit **disjoint** halves of the 23 search sequences, so the joint outputs demonstrably contain queries neither row yields on its own — verified end to end: the joint classification has 187 rows against 105 + 82 for the two individual ones, over 23 distinct queries against 11 + 12.

Placing bacterial sequences on an archaeal reference tree is deliberate and not biologically meaningful — it puts them on long branches. The point is two different profiles legitimately sharing one reference tree, which is the situation the feature exists for, using reference data already present on this branch. No new data files are added; the sheet only references existing ones.

`README.md` documents it under the phyloplace section, including the caveat above.

Generated by Claude

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,19 @@ So the file covers a group with a taxonomy, a group without one, and a row that 
 Nothing else in the file changed, and the column is ignored by pipeline versions that do not know about it.
 Consumed by nf-core/phyloplace's `test_phyloplace_input` profile.
 
+### `testdata/phylosearch_input_reftree.csv` - grouped "search and place" sample sheet
+
+A two-row sample sheet for nf-core/phyloplace's "search and place" mode, in which both rows carry the same `reftreename` and so are summarised jointly as well as individually.
+It exists as a separate file rather than as extra columns on `phylosearch_input.csv`, so that the original sheet keeps covering the layout of a sample sheet written before the `reftreename` column existed.
+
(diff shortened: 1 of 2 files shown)


dialvarezs (approved):
Are the Figshare files too heavy to add them here?

Otherwise, looks good!

erikrikarddaniel:
> Are the Figshare files too heavy to add them here?
> 
Are you referring to the reference trees we use in Ampliseq? I added the archaeal tree here before I had them, so I'm using the files here.

> Otherwise, looks good!

Thanks!

PR #2237, opened 2026-08-25
feat: add metagenome dataset for binning

dialvarezs:
Adds a mock metagenome dataset for binning with 2 species and 3 samples.
More details in the README.

PR #2255, opened 2026-09-02
feat: add per-sample assemblies to the metagenome binning dataset

dialvarezs:
`SemiBin2 multi_easy_bin` and `Vamb`, and co-assembly binners in general, need one assembly per sample combined into a single FASTA with the sample as a header prefix. The binning dataset added in #2237 shares one assembly across its three samples, so it cannot drive that mode, and nf-core/modules still tests `semibin/multieasybin` against data under `delete_me/`.

This adds `binning/multisample/`, covering the same community in that form. It is derived entirely from the files already in `binning/`, so there is no second simulation to keep in step:

- the reads come back out of the published BAMs with `samtools fastq`, split by genome on the reference name
- the contigs are the published 5 kb pieces stitched into blocks of 8, with each sample taking a seeded 60% of them, which keeps the contig count near what a real assembly would give
- each sample takes one genome's reads from one published sample, giving five abundance ratios out of the three that exist

Contents are `contigs.fasta.gz`, `s1` to `s5` `.sorted.bam` with indexes, `s1` to `s5` `.aemb.tsv` and `abundances.tsv`.

It carries five samples rather than three because SemiBin2 only accepts `--abundance` in place of BAMs from five samples up. Both tools' abundance forms are included, since they differ: SemiBin reads one file per sample over the split contigs `SemiBin2 split_contigs` produces, Vamb one wide table over the contigs themselves.

The separator is `C`, which is Vamb's default and what SemiBin takes as `-s C`. Using SemiBin's own default of `:` works for both tools but leaves a colon in the names of the bins Vamb writes.

Verified on all four paths. SemiBin2 2.4.1 returns 2, 2, 3, 2 and 2 bins from the BAMs and 2 per sample from the abundance files, about 70 seconds a run. Vamb 5.0.4 returns 53 clusters in 163 split bins from either input, at the low epoch count nf-core/modules uses for testing. The README section documents the full recipe and the seeds.

Draft until nf-core/modules#12863 is pointed at it.


erikrikarddaniel (approved):
👍

PR #2256, opened 2026-09-02
Reduce archaeal species-preference test data from 7 to 3 species

erikrikarddaniel:
## Description

Reduces the archaeal genome-selection/species-preference test dataset (originally added for nf-core/magmap#187) from 7 species / 14 genomes down to 3 species, to address [nf-core/magmap#246](https://github.com/nf-core/magmap/issues/246): `sourmash_genome_selection.nf.test` and `species_preference.nf.test` were extremely slow because 7-8 selected genomes per test each needed a full real Prokka annotation (no pre-computed GFF is possible for genomes fetched live from NCBI at test time).

The 3 species kept (Metallosphaera javensis, Aeropyrum pernix, Ignisphaera cupida) were chosen specifically to preserve every `--species_preference` (local/completeness/gtdb) and `--genomeset_mode` (joint/sample) behaviour the original 7-species set exercised -- see the new README section for the full reasoning and the actual CheckM/GTDB completeness numbers behind each choice. All 3 reuse genomes, GFFs, and GTDB metadata already in this dataset; nothing new was downloaded from NCBI.

Also fixes a design smell in the original dataset: its three read samples were built from entirely disjoint species, so `--genomeset_mode joint` vs `sample` was never actually exercised against a genome appearing in more than one sample. The three new samples deliberately share species pairwise instead.

`testdata/archaeal_duplicates.gtdbtk.tsv` (all 7 species) is left untouched, since nf-core/magmap's `test_bakta` profile still uses it via `bakta_test.genomes.csv`, which references all 6 `local_*` genomes directly. Everything else from the original 7-species set is removed as now-orphaned.

All matches were verified with real `sourmash sketch`/`prefetch` runs before committing (not just assumed): every sample's simulated reads cleanly match both their public-representative and local-genome signatures, well above the pipeline's Sourmash Gather threshold.

Companion PR on nf-core/magmap switching `test_sourmash_genome_selection`/`test_species_preference` to this new dataset will follow once this merges.

_Generated with some help from Claude Code._


piplus2 (approved):
LGTM

PR #2257, opened 2026-09-03
Add chunks obtained with glimpse v2

LouisLeNezet:
This add GLIMPSE2_chunk output for unittest


get_panel_s.sh
mahesh-panchal, line 1: The presence of shell file on a test-datasets confuses me, but since it's present it's been approved before.

mahesh-panchal (approved):

get_panel_s.sh
LouisLeNezet: The shell file is here only to show which command has been used to generate the test data.
I took inspiration from the modules branch, but as we have many tool ran, I wanted to make it easy to rerun if we wanted to expand to other regions / individuals.

mahesh-panchal: Very good point and practice. I need to remember that.

PR #2259, opened 2026-09-04
Add easypqp convert pickles for easypqp/library module test

jonasscheid:
Adds `proteomics/easypqp/HepG2_rep1_small.{psmpkl,peakpkl}`, the `easypqp convert` output of the HepG2 idXML + mzML pair merged in #2258.

- Input for the `easypqp/library` module test (nf-core/modules#12891).
- Documented in the proteomics README.

### DIFF
--- a/data/proteomics/README.md
+++ b/data/proteomics/README.md
@@ -6,6 +6,7 @@
 - [cif](#cif)
 - [database](#database)
 - [diann](#diann)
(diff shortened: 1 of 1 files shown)


jonasscheid:
Not needed: the easypqp/library module test will chain easypqp/convert via an nf-test setup block instead.

PR #2263, opened 2026-09-08
Add CADD prescored fixtures and generator script for annotate_cadd in raredisease

ramprasadn:
File descriptions are in the README file.


peterpru (approved):
This looks good to me!

scripts/build_cadd_prescored_fixture.py
```diff
@@ -0,0 +1,423 @@
...
+    segments: list[ProvenanceSegment],
+    file_is_chr_prefixed: bool,
+):
+    """tabix every provenance segment of ``local_contig`` out of ``prescored_gz``
+    and yield each row as a :class:`PrescoredVariant` on sliced local
+    coordinates. Rows that fall outside the mapped local window are dropped."""
+    local_contig_nochr = strip_chr(local_contig)
+    for segment in segments:
+        query_chrom = segment.real_chrom if file_is_chr_prefixed else strip_chr(segment.real_chrom)
+        region = f"{query_chrom}:{segment.real_start}-{segment.real_end}"
```
peterpru, line 204: One possible caveat here, the script assumes that querying real_start-real_end is sufficient to ensure the entire variant belongs to that provenance segment. But the filtering only validates the variant's POS, not its complete REF interval. This is completely fine for SNVs, as they should be size 1, but for indels it could theoretically be possible that the end is outside of the segment, while being counted as being in the segment.

However, since this is test data and you have run it with these files, it might not be worth the effort to spend more time on that.

ramprasadn: Good point! Addressed in https://github.com/nf-core/test-datasets/pull/2264

PR #2264, opened 2026-09-08
Harden CADD fixture builder: check full REF span against segment bounds

ramprasadn:
Follow-up to #2263, addressing review comment https://github.com/nf-core/test-datasets/pull/2263#discussion_r3957492349.

`remap_prescored_contig` filtered segment membership on POS only, so a multi-base indel near a segment boundary could be lifted onto local coordinates that partly belong to the adjacent segment or the unmapped gap between chr20's two segments. The check now requires the whole REF span (POS .. POS + len(REF) - 1) to fall inside the segment.


peterpru (approved):
Great!

PR #2267, opened 2026-09-11
Add sourmash test database and update database CSV

haris18s:
Add sourmash test database (contains sourmash-db.zip + lineages.csv.gz as proposed in previous  [PR](https://github.com/nf-core/taxprofiler/pull/404))


jfy133 (changes requested):
You need to make a whole new sample sheet for 3.1 ,as this one will fail runs of the current version as sourmash will not be considered a valid input

jfy133 (approved):

PR #2268, opened 2026-09-11
Add chr prefix to gnomAD.r2.1.1-sv.vcf.gz for svdb/query test

ramprasadn:
## Summary

`gnomAD.r2.1.1-sv.vcf.gz` used bare chromosome names (`22`) while `sv_query.vcf.gz` (the query VCF used against it in `nf-core/modules`' `svdb/query` test) uses `chr`-prefixed names (`chr22`). This mismatch means the two files describe the same genomic positions but never lexically match.

svdb <= 2.8.4 silently stripped `chr` prefixes on read, so the mismatch didn't matter in practice. svdb 2.9+ intentionally removed that normalization, so with svdb 2.12.0 the two files no longer overlap at all, and the `svdb/query` test's only annotated hit silently disappeared.

This PR fixes that by adding `chr` prefix to `gnomAD.r2.1.1-sv.vcf.gz`'s CHROM column, `##contig` header IDs, and `CHR2` INFO values so it's consistent with `sv_query.vcf.gz`, restoring the intended annotation match under current and future svdb versions.

This file is only referenced by `nf-core/modules`' `svdb/query` test, so this change is isolated.


beatrizsavinhas (approved):
Very nice description of the investigation! And great that you already checked that the file is not used for anything else too! 🌟

PR #2276, opened 2026-09-17
Funcprofiler: fix fulltest databbase and samplesheet

nickp60:
Add mifaser to full test, add copy of taxprofiler's samplesheet omitting samples that contained both single-end and paired libraries.


miraep8 (approved):
Looks good to me!  Thank you!  Small suggestion if you really want to lock down the exact version of these files to consider using the commitment hash `cf59872f190e99b06b3e6b77f00f57acc76d18e4` instead of `taxprofiler` in the url. But I think those files are meant to be static, so I also don't anticipate it being an issue if you don't do this.

PR #2280, opened 2026-09-18
Add salmon 2.7.0 pre-built index fixture for rnaseq

pinin4fjords:
## Summary
- salmon 2.7.0 (the Rust rewrite) uses an incompatible index format, so `salmon quant` can no longer read the existing `reference/salmon.tar.gz` fixture used by nf-core/rnaseq's test profile.
- Adds `reference/salmon_2.7.0.tar.gz`, built from the same genome/GTF/spike-in test data via the pipeline's own `SALMON_INDEX` process (genome-as-decoy, transcriptome regenerated from `--fasta`/`--gtf`/`--additional_fasta`), alongside the existing fixture rather than replacing it.
- Verified end-to-end: a full nf-core/rnaseq run using this file as `--salmon_index` completes successfully with real salmon 2.7.0 quantification.

## Test plan
- [x] Built fixture via `nextflow run nf-core/rnaseq -profile test,docker --salmon_index null --transcript_fasta null --save_reference true`
- [x] Verified `--salmon_index reference/salmon_2.7.0.tar.gz` completes a full pipeline run successfully


erikrikarddaniel (approved):
👍

PR #2282, opened 2026-09-20
Publish PLINK bundles and drop the retired plink2 selector on the gwas branch

lyh970817:
Follow-up to #2198 for the `gwas` branch, so nf-core/gwas can run `-profile test` directly from the published data.

- Drops `plink2` from the association selectors in the four analysis manifests: nf-core/gwas removed its PLINK 2 `--glm` route, so that token is no longer an accepted `association_methods` value. The generator and validator are aligned so regeneration reproduces the CSVs byte for byte.
- Publishes the PLINK 2 (`example_all.pgen/.psam/.pvar`) and PLINK 1 (`example_all.bed/.bim/.fam`) representations of the compact two-chromosome VCF, derived in the workflow with plink2 2.0.0a.6.9 (`--double-id --make-pgen`; `--make-bed --hard-call-threshold 0.1`). The regeneration test re-derives them and asserts byte identity, as it already did for the VCF.
- Switches the cohort manifest to select the PLINK 2 bundle by its public URLs; the pipeline's cohort schema takes PLINK bundles, not VCF. The VCF stays the canonical source.
- README updated accordingly.

20 files under `results/fixtures`, unchanged sizes for everything already on the branch.


chriswyatt1:
Great, I will go over this properly on Tuesday next week. We have been a little busy with a paper submission, but finally have time :)

lyh970817:
> Great, I will go over this properly on Tuesday next week. We have been a little busy with a paper submission, but finally have time :)

Thanks! This is just to make the test datasets aligned with my current pipeline repo which has retired the VCF route and an association method.

chriswyatt1:
Hey @lyh970817 , have the results been accidentely moved into this branch, or did you want them here?

modules/validate_gwas_fixtures.nf
```diff
@@ -0,0 +1,38 @@
...
+	input:
+	path vcf
+	path pheno
+	path qcovar
+	path catcovar
+	path relational_dir
+	tuple path(pgen), path(psam), path(pvar)
+	tuple path(bed), path(bim), path(fam)
+
+	output:
```
chriswyatt1: On these local modules, they need to have a versions.yml 
Plus this should be in a local folder, not in the main module dir. Same for the rest, there should be distinction between nf-core ones and local.

Ideally we should also add stub and unit test etc. But we could leave this on the roadmap, if you just want to check it all works as expected.

chriswyatt1: We should try to make them as nf-core compliant as possible a this stage really, with correct container lines as well.

chriswyatt1, line 15: use of meta.id as well,,, probably. I need to look into the module

chriswyatt1 (changes requested):

modules/validate_gwas_fixtures.nf
chriswyatt1: Maybe its not necessary for test datasets though, i will check!

chriswyatt1: I see, yes, this is fine as is, just wanted to be sure this was ok :)

chriswyatt1 (approved):
Great, this seems good to go

PR #2284, opened 2026-09-21
Add UCSC-named copies of the chr4/chr19 variant fixtures

axelwalter:
Every variant fixture on this branch uses Ensembl contig names (`4`, `19`). A pipeline that renames UCSC names to Ensembl ones therefore has that step exercised only as a no-op, and a rename bug passes CI unnoticed. Real callers emit UCSC names: Mutect2 and HaplotypeCaller both write `chr1`, `chr4`, `chrM`.

This came out of nf-core/epitopeprediction#362, where germline context was silently dropped for every real user because the germline VCF kept its `chr` prefix while the somatic calls had already been renamed. The existing fixtures could not catch it; a real exome pair did.

### Files

- `testdata/variants/variants_chr4_chr19_ucsc.vcf` — `variants_chr4_chr19.vcf` with `chr`-prefixed contigs.
- `testdata/variants/variants_chr4_chr19_germline_ucsc.vcf` — `variants_chr4_chr19_germline.vcf` with `chr`-prefixed contigs.
- `testdata/sample_sheets/sample_sheet_variants_grch38_ucsc.csv` — pairs the two.

Only the contig names change, in both the `##contig` headers and the records. Same four somatic and two germline records at the same positions, everything else byte-identical.

### Why this shape

The reference and VEP cache on this branch stay Ensembl-named, which is the real-world arrangement: callers emit UCSC names, Ensembl resources do not, and the pipeline bridges them. Because only the naming differs, a correctly renaming pipeline produces exactly the results the Ensembl-named fixtures produce, so a test can assert the same expected peptides and any difference is a rename failure.


axelwalter:
Closing in favour of a simpler change: rather than adding parallel UCSC-named copies, the germline fixture itself should be chr-prefixed by default. It is brand new and used by one test, real germline callers (HaplotypeCaller, Strelka) emit chr-prefixed contigs anyway, and leaving the somatic fixture Ensembl-named reproduces the mixed-naming case that actually breaks pipelines. No new files, no new test: the existing germline test then covers the rename, and its expected output is unchanged because renaming is meant to be transparent. Replacement PR to follow.

PR #2288, opened 2026-09-23
Simplify vcf header and SNP extraction

LouisLeNezet:
- Replace plink SNP position extraction with `bfctools view -R ${SNP_FILE}` option.
- Remove unseen contigs from vcf header (keep only chr21 and chr22).
- `ind.s.bcf` replaced by `ind.s.vcf.gz` for easier debugging (`zcat` usable directly).
- Merge 3 samples vcf in one.
- Remove `ind.s.1x.bcf` as only used as temporary files for glimpse1.


erikrikarddaniel (approved):
👍

